Issue #25, opened 2019-01-14
Flux Models


MikeInnes:
This package looks cool and I wonder what it'd take to get it working with Flux models. I'm basically imagining replacing 
```julia
nnet = read_nnet("examples/networks/small_nnet.nnet")
```
with
```julia
nnet = Chain(Dense(...), ...)
```
and passing it to `Problem(nnet, input_set, output_set)` still works.

Under the hood it should be a simple case of converting Flux chains to this package's internal format, something I'm happy to help out with if there's interest.

mykelk:
Awesome! There is definitely interest in this. I think you can bypass the `.nnet` file entirely and just populate the `Network` structure in `src/utils/network.jl`. It is super simple; just a vector of weight matrices and bias vectors along with an activation function.

tomerarnon:
Converting from `Flux`'s representation to ours would be quite simple assuming only `Dense` layers. For function layers or user-defined layers, this can get much hairier, of course.
```julia
import NeuralVerification: Layer, ReLU

Layer(d::Dense) = Layer(d.W, d.b, d.σ)
function Network(C::Chain)
    @assert all(isa.(C.layers, Dense)) "NeuralVerification only supports Dense layers"
    Network([Layer(d) for d in C.layers])
end

# e.g.:
julia> d = Dense(ones(1,1), ones(1), ReLU());

julia> Network(Chain(d,d))
Network(Layer[Layer([1.0], [1.0], ReLU()), Layer([1.0], [1.0], ReLU())])
```
Passing along `Chain`s directly (without conversion) could also be possible actually, but is much more involved. It would be interesting if at some point we could generalize the idea of a `Layer` so that it could work with `Dense` (or any others) directly, and potentially even leverage Flux's GPU capabilities as a result...

But we're not yet ripe enough to consider that too seriously!

PR #35, opened 2019-01-25
Flux conversions

MikeInnes:
As there was interest in #25 I've added some simple conversion utilities here. It's bidirectional so you can do things like:

```julia
julia> using NeuralVerification, Flux

julia> nnet = Chain(read_nnet("examples/networks/small_nnet.nnet"))
Chain(Dense(1, 2, NNlib.relu), Dense(2, 2, NNlib.relu), Dense(2, 1, NNlib.relu))

julia> nnet([1])
1-element Array{Float64,1}:
 78.5

julia> input_set  = Hyperrectangle(low = [-1.0], high = [1.0])
julia> output_set = Hyperrectangle(low = [-1.0], high = [70.0])
julia> problem = Problem(nnet, input_set, output_set)

julia> solve(BaB(), problem)
CounterExampleResult(:UNSAT, [1.0])
```

If the approach seems right I'm happy to add appropriate tests etc.


MikeInnes:
Sure thing, I've set this up to use requires instead of a direct dependency.

tomerarnon:
@MikeInnes, thanks again for setting this up.
I added a couple of basic tests for this functionality, but since Flux is an optional dependency, I am having a hard time actually getting the tests to run. I noticed your call to `Base.find_package` in [Flux's tests](https://github.com/FluxML/Flux.jl/blob/master/test/runtests.jl) so I tried imitating that in ours, but of course Flux isn't being added for the build since it's not in the Project file so the tests just never get called. I couldn't find anything special in Flux's .travis file that could remedy this, but I'm assuming there's a way to get it to work. Whats the secret sauce?

MikeInnes:
Right, that test only works if CuArrays is available in the global environment. CUDA isn't available on Travis so they actually only kick on our GPU CI, which has CuArrays installed globally.

Your options are basically (a) use `find_package` and only test if/when Flux installed (maybe on your local machine) or (b) just add Flux to the test dependencies so that it will get installed on CI.

PR #50, opened 2019-02-27
Mixed networks

clazarus:
- Reluplex now supports mixed layers
- all tests appear to pass (2 broken tests in relu_networks)
- small change to aux function in variables.jl (appears to only be called by Reluplex)
- adding inactive_relus.jl to test correct functionality of problems where ReLUs are always inactive


tomerarnon:
Looks good. There's one point I've never liked about our implementation which is that the first layer's variables are "backwards facing". What would be the consequences of categorizing them as forward facing? It would make our `if (i==1)` branch go away for one, which is good. But I don't know how it would affect other matters. I think it would make more sense conceptually. Any thoughts?

Also, at the end of this pull, it would be good to keep only one of the two relu testsets and ensuring the test region is nontrivial.

changliuliu:
Maybe we can assign the input layer an ID activation, then we can have both forward facing and backward facing variables for the input layer.

tomerarnon:
Not sure if this is how it would actually look, but it makes sense to me in theory. Without changing anything else in the code:

```julia
function encode(solver::Reluplex, model::Model,  problem::Problem)
    layers = problem.network.layers
    bs = init_variables(model, layers)
    fs = init_variables(model, layers, include_input = true)

    bounds = get_bounds(problem)
    for (i, L) in enumerate(layers)
        add_set_constraint!(model, bounds[i], bs[i])
        W, bias = L.weights, L.bias

        @constraint(model, W*fs[i] .- bs[i+1] .== -bias)

        intralayer_constraint!(model, fs[i], bs[i], L.activation)
    end
    add_complementary_set_constraint!(model, problem.output, last(fs))
    zero_objective!(model)
    return bs, fs
end
```

The idea being that now `fs` is of length L+1 instead of `bs` and the "interlayer constraint" depends on `Bᵏ⁺¹ and Fᵏ` and the "intralayer constraint" depends on `Bᵏ and Fᵏ`

I also just glanced at the reluplex section of the manuscript and saw the Bᵏ, etc., notation, which we may want to adopt here, assuming it carries over naturally.

Does this look right to both of you?

tomerarnon:
@clazarus before merging, would be good to clean up/finalize all of the new code. 

Modifying `Reluplex` as proposed is optional but if it's correct and desired I can go ahead and do it.

clazarus:
Input variables where originally treated as backward facing (b) because b variables have no intrinsic constraints on their ranges, whereas forward (f) variables where constrained to always be non-negative (due to ReLU's) and this is how they are treated in the original Reluplex paper and code.

I believe adding an Id() layer for the input variables would add unnecessary equations and variables to every problem.

I see the aesthetic gain in making the input variables fs but it does make their intended meaning/origin depart from their initial purpose which is to help encode ReLU's. I will have a careful look at the code and see where my vote goes.

tomerarnon:
I just pushed a couple of commits that update Reluplex as proposed. I also aligned our notation with that used in the paper (for better or worse, you be the judge).

In particular take a look at `encode` of course, but also note that introducing input layer forward variables eliminates some `[i+1]`s in other places in the code.

clazarus:
I agree that the changes to Reluplex made it more aesthetic and easy to read. I have to say that adding an Id layer for the input layer creates as many additional variables and equations (constraints) as the dimension of the network. I think this would be undesirable if performance was the only goal but I think the primary purpose of this implementation is pedagogical so I think we should stick to it. I have also rerun the performance tests and noticed no decrease after this change. I will wait for Travis to finish and merge this PR.

PR #52, opened 2019-03-03
Complements

### LINKED ISSUE #44: Encode some output constraints as "PolytopeComplement"
As per the discussion earlier today, we should create a new type which represents the complement of a polytope. This will allow us to capture more expressive output sets in certain algorithms.

tomerarnon:
Mostly closes #44. Note that this only implements `PolytopeComplement`s for optimization methods that require it. Other methods do not support it due to issues discussed in #44.

Next, we should set up an interface for checking valid problem/solver pairings since some solvers only support some set types, and `PolytopeComplement`s only complicate this further. `Problem` is parameterized by the set types it contains, so it is not complicated to make e.g. `solve(MaxSens::DLV, problem::Problem{In, Out})` but unfortunately this will be unacceptably ugly e.g.` Problem{Union{Hyperrectangle, HPolytope, PolytopeComplement}, Union{...}}`.

Right now, my best idea is to define methods such as 
```julia 
function iscompatible(solver::MaxSens, problem::Problem)
    typeof(problem.input)  ∈ (Hyperrectangle, HPolytope, PolytopeComplement) # etc.
    typeof(problem.output) ∈ (Hyperrectangle, HalfSpace, HPolytope, PolytopeComplement) # etc.
end
```
For each solver, but I don't like that very much either. Alternatively defining `inputsets(solver)` and `outputsets(solver)` to achieve the same thing.

This has come up once before already, but since it hasn't yet been resolved, ideas are still welcome.


changliuliu:
I like the ```iscompatible``` method.

src/utils/problem.jl
```diff
@@ -3,15 +3,14 @@
     PolytopeComplement
 
 The complement to a given set. Note that with the exceptions of `Halfspace` and `EmptySet`,
- a `PolytopeComplement` is not a convex set.
-The.
+ a `PolytopeComplement` is not a convex set. Also note that `PolytopeComplement`s are open by definition.
```
changliuliu: better to say: a `PolytopeComplement` may not be a convex set. 
For half space, the complement is convex.

docs/src/problem.md
```diff
@@ -11,6 +11,60 @@
...
+ - HP = `HPolytope`
+ - PC = `PolytopeComplement`
+
+[](TODO: review these. Not sure they're all correct.)
+
+|        Solver        |  Input set  |    Output set    |
+|----------------------|:-----------:|:----------------:|
+| [`ExactReach`](@ref) | HP          | HP (bounded[^1]) |
+| [`AI2`](@ref)        | HP          | HP (bounded[^1]) |
+| [`MaxSens`](@ref)    | HP          | HP (bounded[^1]) |
```
changliuliu: I suppose MaxSens takes both HP and HR. Or are we using HP as a more general object that includes HR?

docs/src/problem.md
```diff
@@ -11,6 +11,60 @@
...
+| [`NSVerify`](@ref)   | HR          | PC[^2]           |
```
changliuliu: NSVerify can take both HR and HP for input.

src/utils/problem.jl
tomerarnon: > For half space, the complement is convex.

This is what I was trying to say with:

>Note that with the exceptions of `Halfspace` and `EmptySet`, a `PolytopeComplement` is not a convex set.

but I guess it didn't get across as I intended

docs/src/problem.md
tomerarnon: This is a distinction that I was struggling with as well. I put a note about it after the table, but I don't consider the point resolved... We may end up going with HP/HR/HS for each one that takes multiple inputs. The additional issue with solvers like MaxSens is that actually, for HP input, it _overapproximates_ as HR anyway, so saying it takes HP is only true with a caveat.

src/utils/problem.jl
schillic: Not sure if you use that, but the complement of an `EmptySet` is a `Universe`, so that would need to be added to the list.

tomerarnon: @schillic luckily then, we don't use it 😄. We don't actually use the `EmptySet` either, it just occurred to me as another exception to the convexity statement.

docs/src/problem.md
```diff
@@ -11,6 +11,60 @@
...
+
+ [^3]: The set can only have one output node. I.e. it must be a set of dimension 1.
+
+Note that solvers which require `Hyperrectangle`s also work on `HPolytope`s by overapproximating the input set. This is true for any set which can be converted to another set, such as a `HalfSpace` output being representable as an `HPolytope` with a single constraint.
+
+### PolytopeComplements
+
+Some optimization-based solvers work on the principle of a complementary output constraint.
+In essense, they test whether a point *is not* in a set, by checking whether it is in the complement of the set (or vice versa).
+To represent the kinds of sets we are interested in for these solvers, we define the `PolytopeComplement`, which represents the complement of a convex set.
```
clazarus: I would remove "convex" from: "which represents the complement of a convex set"

```suggestion
To represent the kinds of sets we are interested in for these solvers, we define the `PolytopeComplement`, which represents the complement of the output set.
```

PolytopeComplements do not necessarily have to represent the complement of a convex set. I think the only connection to convexity is that the HPolytope object can only represent sets define by the conjunction of linear constraints (intersections of half-spaces) which are coincidentally convex. This means that even PolytopeComplements will always be convex but it does not imply that they are the complement of a convex set. As a counter example imagine the unit square in 2 dimensions. It could be a valid PolytopeComplement but its complement is not convex (the whole of the plane except for a square hole missing around the origin).

docs/src/problem.md
```diff
@@ -11,6 +11,60 @@
...
+ [^2]: See [`PolytopeComplement`](@ref) for a justification of this output set restriction.
+
+ [^3]: The set can only have one output node. I.e. it must be a set of dimension 1.
+
+Note that solvers which require `Hyperrectangle`s also work on `HPolytope`s by overapproximating the input set. This is true for any set which can be converted to another set, such as a `HalfSpace` output being representable as an `HPolytope` with a single constraint.
+
+### PolytopeComplements
+
+Some optimization-based solvers work on the principle of a complementary output constraint.
+In essense, they test whether a point *is not* in a set, by checking whether it is in the complement of the set (or vice versa).
```
clazarus:
```suggestion
In essence, they test whether a point *is not* in a set, by checking whether it is in the complement of the set (or vice versa).
```

docs/src/problem.md
tomerarnon: Indeed. However, we define PolytopeComplements as `PolytopeComplement{S<:LazySet}` and LazySets do not represent non-convex sets, so it's still sort of true. I see your point though. I don't want to necessarily tie the concept of a `PolytopeComplement` to the output set, even though that is what it was made for. It can be more general than that ultimately.

clazarus: But the fact that an object is a convex set does not imply that it is the complement of a convex set. Again, think of the unit square, it is a convex set but it is the complement of a non-convex set.

tomerarnon: Isn't that what the paragraph says? 
>represents the complement of a convex set.

No mention of PCs being convex themselves. And the next sentence:
>Note that in general, the complement of a convex set is neither convex nor closed.

Rereading your first comment:
>As a counter example imagine the unit square in 2 dimensions. It could be a valid PolytopeComplement but its complement is not convex (the whole of the plane except for a square hole missing around the origin).

That could not be a valid PC. PC is specifically defined as a lazy complement to a LazySet, and therefore could not be a unit square. Because they are specifically the complements of `<:LazySet`, they are effectively also the complements of convex sets by definition (I don't think LazySets deals with any nonconvex sets.)

It may be worth rethinking the name though, since `<:LazySet` includes sets other than polytopes, which could serve as perfectly good complements also.

schillic:
FYI: We will add a `Complement` type in `LazySets` soon.
However, we cannot provide all your methods there.

https://github.com/sisl/NeuralVerification.jl/blob/93c6581c6cfd08b1f2e835d1111e9ceac500b39b/src/utils/problem.jl#L28-L36

For instance, this line is generally incorrect.

https://github.com/sisl/NeuralVerification.jl/blob/93c6581c6cfd08b1f2e835d1111e9ceac500b39b/src/utils/problem.jl#L30

Counterexample: Let `C1`, `C2` be non-intersecting circles. Then the complement of `C1` is not contained in `C2` (in particular, the complement is not even bounded). The other way around it works, though (`C2` is contained in the complement of `C1`).

circles

I did not check how you restrict the types of your `PolytopeComplement`, but maybe you want to double-check that everything is sound here.

changliuliu:
@chris Great to know that you will support Complement type!

Yes, this line is incorrect.
[NeuralVerification.jl/src/utils/problem.jl](https://github.com/sisl/NeuralVerification.jl/blob/93c6581c6cfd08b1f2e835d1111e9ceac500b39b/src/utils/problem.jl#L30)
Line 30 in [93c6581](/sisl/NeuralVerification.jl/commit/93c6581c6cfd08b1f2e835d1111e9ceac500b39b)
```  LazySets.issubset(PC::PolytopeComplement, s) = LazySets.is_intersection_empty(s, PC.P) ```

We actually do not need all functions that are currently there. For now, as discussed with @tomerarnon, we are only using PolytopeComplement in optimization-based methods. Those supporting functions `issubset` `is_intersection_empty` are needed for reachability-based methods. Will update the code to remove inaccurate or unnecessary functions.

File: src/utils/problem.jl [hunk 6]
```diff
@@ -1,15 +1,51 @@
...
+true
+```
+"""
+struct PolytopeComplement{S<:LazySet}
+    P::S
+end
+
+Base.show(io::IO, PC::PolytopeComplement) = (println(io, "PolytopeComplement of:"), println(io, "  ", PC.P))
+LazySets.issubset(s, PC::PolytopeComplement) = LazySets.is_intersection_empty(s, PC.P)
+LazySets.issubset(PC::PolytopeComplement, s) = LazySets.is_intersection_empty(s, PC.P)
```
changliuliu: Let's remove this line.

tomerarnon: Done. Thanks @schillic for pointing this out! That method was never called of course, but now it is also gone! Excited to adopt `LazySets`' `Complement` type and have the pros handle it 😉

schillic: :wink:
The second method you removed was actually fine (`is_intersection_empty` is symmetric in its arguments).

tomerarnon: Indeed, but they were both unused, and the PC type will be short lived anyway. 
They were added together and removed together...

Issue #71, opened 2019-09-13
JuMP return code mishandling leading to inconsistency


tomerarnon:
I think the cause of #70 is that we're mishandling some `JuMP` return codes. We test for `OPTIMAL` and `INFEASIBLE`, but we are sometimes also seeing `INFEASIBLE_OR_UNBOUNDED`. 

First we should consider whether any of the objectives we use can ever by unbounded in the first place. If not, we can always treat the problems as infeasible in these cases. If it _is_ possible, then we need to consider what to do in unbounded cases. E.g., consider MIPVerify:

```julia
##########################################
# currently
o = objective_value(model)
if termination_status(model) == INFEASIBLE
    return AdversarialResult(:holds)
end
if o >= maximum(problem.input.radius)
    return AdversarialResult(:holds)
else
    return AdversarialResult(:violated, o)
end
##########################################


# fixed?
o = objective_value(model)
if isfeasible(model) && o < maximum(problem.input.radius)
    return AdversarialResult(:violated, o)
else
    return AdversarialResult(:holds, o)
end
```
(Note that `isfeasible` is some custom function that tests for feasibility according to our criteria, solving a feasibility problem if necessary).

In the case where `model` is feasible but unbounded, we would never want to still return `o` as an adversarial result, given that it's NaN. 
In this particular case we're safe from that, since `NaN < some_num == false`, but in other solvers this may not be as cut and dry.

What about DLV:
```julia
# formerly: if termination_status(model) == OPTIMAL
if isfeasible(model)
    return (true, value(first(neurons)))
else
    return (false, [])
end
```
In the unbounded (but feasible) case, can we really still return `value(first(neurons)))` as the counterexample (it will probably be full of NaNs)? If not, given that the problem is indeed violated, what can we return instead?

Again, not sure if the problems we're dealing with can ever _actually_ be unbounded or whether that return code always applies to infeasibility. @changliuliu, I could definitely use your input on this! I'll be opening a PR shortly that begins to address this. Hopefully we can integrate it before next week.

tomerarnon:
Upon second review, here's what I've come up with:

DLV, iLP, MIPVerify use `max_disturbance!`, which minimizes the infinity norm, and is therefore naturally bounded. MIPVerify, which _was_ affected by this bug, was affected because it was explicitly testing `INFEASIBLE`. If it had explicitly tested `OPTIMAL` instead, it would have been safe. This is a simple change to make in these three cases.

[...]

PR #72, opened 2019-09-13
Fix opt

tomerarnon:
Opening this PR that addresses #71 so that we can review and discuss the changes. It is not yet ready to merge.

### DIFF
--- a/src/optimization/certify.jl
+++ b/src/optimization/certify.jl
@@ -47,19 +47,24 @@ function solve(solver::Certify, problem::Problem)
     @constraint(model, diag(P) .<= ones(n))
     @objective(model, Max, first(o))
     optimize!(model)
-    return interpret_result(solver, termination_status(model), first(o))
-end
-
-# True if o < 0
-# Undertermined if otherwise
-function interpret_result(solver::Certify, status, o)
-    # println("Upper bound: ", value(o[1]))
     if value(o) <= 0
         return BasicResult(:holds)
     else
         return BasicResult(:unknown)
     end
+    # return interpret_result(solver, termination_status(model), first(o))
 end
+
+# True if o < 0
+# Undertermined if otherwise
+# function interpret_result(solver::Certify, status, o)
+#     # println("Upper bound: ", value(o[1]))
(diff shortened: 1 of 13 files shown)


src/optimization/certify.jl
```diff
@@ -47,20 +47,25 @@ function solve(solver::Certify, problem::Problem)
...
-# True if o < 0
-# Undertermined if otherwise
-function interpret_result(solver::Certify, status, o)
-    # println("Upper bound: ", value(o[1]))
     if value(o) <= 0
         return BasicResult(:holds)
     else
         return BasicResult(:unknown)
     end
+    # return interpret_result(solver, termination_status(model), first(o))
```
tomerarnon: `termination_status` wasn't being used here.

src/utils/util.jl
```diff
@@ -350,4 +350,42 @@ function split_interval(dom::Hyperrectangle, i::Int64)
...
     input_split_right = Hyperrectangle(low = input_lower, high = input_upper)
     return (input_split_left, input_split_right)
+end
+
+
+"""
+    isfeasible(model::Model)
+Test if the termination status of `model` can be considered feasible.
+"""
+function isfeasible(model::Model)
```
tomerarnon: This is the crux of the change. `isfeasible`replaces checks like `termination_status(model) == OPTIMAL` to capture the cases we expect to encounter, which may be more diverse.

src/optimization/iLP.jl
```diff
@@ -44,30 +44,41 @@ function solve(solver::ILP, problem::Problem)
     if !solver.iterative
         encode_network!(model, nnet, neurons, δ, StandardLP())
         optimize!(model)
-        termination_status(model) != OPTIMAL && return AdversarialResult(:unknown)
-        return interpret_result(solver, value(o), problem.input)
+        J = value(o)
+        if isfeasible(model)
```
tomerarnon: Upon second review, `iLP` seems to be safe from this issue. Its objective is minimizing the infinity-norm, so is naturally bounded.

tomerarnon:
Closed by #168

PR #79, opened 2019-11-10
Neurify

changliuliu:
implementation of neurify


mykelk:
This is beautiful.

clazarus:
Changliu,
Apologies for the delay. Neurify is failing identity_network.jl. I've investigated the problem and couldn't solve it, Julia crashes without a trace on my local machine. On travis I did get a trace for StackOverflowError.

Given that the only test that fails comes from identity_network.jl I am assuming that the source of the problem is:
https://github.com/sisl/NeuralVerification.jl/blob/debaf5c96bacc6bdfdb5095b16f2f40ef777b81d/src/adversarial/neurify.jl#L195-L201 because I think this version of the function is only called in this test.

I have also modified the test files so that they can be run independently without having to do manual imports, I know some lines are now duplicated across files but I think it makes debugging a lot easier.

changliuliu:
Ummm, I just tested Neurify on identity_network.jl. It passed on my computer. Have you tested it on your computer?

tomerarnon:
@clazarus I haven't tested this myself but the issue can't be here. There's nothing in that function that could cause a stack overflow.

99% chance that this is package related (lazysets probably since this isn't an optimization method) since travis (and presumably also Chris) is seeing it and @changliuliu is not. In particular, running `]test NeuralVerification`, although it takes way longer than running with `include`, will give you the "travis" result, since it will necessarily use the package versions specified by NV even if the ones you normally use are different. E.g. Changliu you might have an older LazySets version and in whatever update we pinned to, some operation that used to be safe became unsafe, or something like that.

clazarus:
@changliuliu I tested it on my computer and it fails both here and on Travis. I just ran the tests as @tomerarnon suggested (`]test NeuralVerification`) and have attached the trace below. 

It does appear to be related to be related to LazySets as calls to Polyhedra functions appear in the trace. I have `LazySets v"1.10.0"` do you have the same version locally?

`Group 5: Error During Test at /Users/USERNAME/.julia/dev/NeuralVerification/test/identity_network.jl:69
  Got exception outside of a @test
  StackOverflowError:
  Stacktrace:
   [1] similar at ./boot.jl:394 [inlined]
   [2] similar at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/adjtrans.jl:144 [inlined]
   [3] Type at ./array.jl:498 [inlined]
   [4] Polyhedra.MixedMatVRep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}, ::Array{Any,2}, ::BitSet) at /Users/USERNAME/.julia/packages/Polyhedra/p4Wnz/src/matrep.jl:111 (repeats 43604 times)
   [5] vrep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}, ::Array{Any,2}, ::BitSet) at /Users/USERNAME/.julia/packages/Polyhedra/p4Wnz/src/matrep.jl:79 (repeats 2 times)
   [6] vrep(::LinearAlgebra.Adjoint{Any,Array{Any,1}}) at /Users/USERNAME/.julia/packages/Polyhedra/p4Wnz/src/matrep.jl:80
   [7] #polyhedron#226(::CDDLib.Library, ::Function, ::VPolytope{Float64}) at /Users/USERNAME/.julia/packages/LazySets/tmJVq/src/VPolytope.jl:507
   [8] #polyhedron at ./none:0 [inlined]
   [9] #convex_hull#116(::CDDLib.Library, ::Function, ::VPolytope{Float64}, ::VPolytope{Float64}) at /Users/USERNAME/.julia/packages/LazySets/tmJVq/src/VPolytope.jl:342
   [10] #convex_hull at ./none:0 [inlined]
   [11] #convex_hull#40(::CDDLib.Library, ::Function, ::Array{VPolytope{Float64},1}) at /Users/USERNAME/.julia/dev/NeuralVerification/src/reachability/ai2.jl:44
   [12] convex_hull(::Array{VPolytope{Float64},1}) at /Users/USERNAME/.julia/dev/NeuralVerification/src/reachability/ai2.jl:42
   [13] symbol_to_concrete(::NeuralVerification.SymbolicInterval{HPolytope{Float64}}) at /Users/USERNAME/.julia/dev/NeuralVerification/src/adversarial/neurify.jl:77
   [14] check_inclusion(::NeuralVerification.SymbolicInterval{HPolytope{Float64}}, ::Hyperrectangle{Float64}, ::Network) at /Users/USERNAME/.julia/dev/NeuralVerification/src/adversarial/neurify.jl:81
   [15] solve(::Neurify, ::Problem{Hyperrectangle{Float64},Hyperrectangle{Float64}}) at /Users/USERNAME/.julia/dev/NeuralVerification/src/adversarial/neurify.jl:58
   [16] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1071 [inlined]
   [17] macro expansion at /Users/USERNAME/.julia/dev/NeuralVerification/test/identity_network.jl:41 [inlined]
   [18] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [19] macro expansion at /Users/USERNAME/.julia/dev/NeuralVerification/test/identity_network.jl:33 [inlined]
   [20] macro expansion at /Users/osx/buildbot/slave/package_osx64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
   [21] top-level scope at /Users/USERNAME/.julia/dev/NeuralVerification/test/identity_network.jl:19
   [22] include at ./boot.jl:317 [inlined]
   [23] include_relative(::Module, ::String) at ./loading.jl:1044
   [24] include(::Module, ::String) at ./sysimg.jl:29
   [25] include(::String) at ./client.jl:392
   [26] top-level scope at none:0
   [27] include at ./boot.jl:317 [inlined]
   [28] include_relative(::Module, ::String) at ./loading.jl:1044
   [29] include(::Module, ::String) at ./sysimg.jl:29
   [30] include(::String) at ./client.jl:392
   [31] top-level scope at none:0
   [32] eval(::Module, ::Any) at ./boot.jl:319
   [33] macro expansion at ./logging.jl:317 [inlined]
   [34] exec_options(::Base.JLOptions) at ./client.jl:219
   [35] _start() at ./client.jl:425`

changliuliu:
@clazarus @tomerarnon My LazySets is actually v1.12.0.

tomerarnon:
Looks like we are pinned to `v1.11.0` here and in master. Updating LazySets was on my todo, so by now we expected it to be the latest version and avoid things like this... But anyway incrementing to `v1.12` on this branch should be safe (no breaking changes for us it seems) and will make Neurify pass.

changliuliu:
@tomerarnon My student Tianhao created a new pull request #95 for Neurify. Is there a way to merge or link this pull request with #95 ?

tomerarnon:
Yes, I think there is. Let me see if I can edit the target branch of that PR to here.

tomerarnon:
Merged #95 into this branch. Before merging this into master, can we devise a test for Neurify beyond the `@no_error` one? That way this PR is completely self contained.
Also, I promise I'll do a more thorough review before the merge like I said I would, but also I trust all the eyes that have already looked at this!

Wei-TianHao:
Hi  @tomerarnon , I am trying to optimize the code to accelerate current implementation. I would appreciate it if you can review it later.
As for the test, I actually tested Neurify on MNIST together with ReluVal on about 10,000 images. I can add the code to the repo as a new runtest later.

tomerarnon:
> Hi @tomerarnon , I am trying to optimize the code to accelerate current implementation. I would appreciate it if you can review it later.

Will do.

> As for the test, I actually tested Neurify on MNIST together with ReluVal on about 10,000 images. I can add the code to the repo as a new runtest later.

It would be _fantastic_ to add that to the repo! Whenever you get to it is fine; as you can tell, my attention to things here is unfortunately sporadic right now, so I'm glad you're on this.

changliuliu:
I got the following error when calling SymbolicInterval. (Julia 1.4, LazySets v1.37.5)

> MethodError: no method matching NeuralVerification.SymbolicInterval(::Array{Float64,2}, ::Array{Float64,2}, ::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}})
>   Closest candidates are:
>     NeuralVerification.SymbolicInterval(::Array{Float64,2}, ::Array{Float64,2}) at /Users/changliuliu/.julia/packages/NeuralVerification/1cnod/src/adversarial/neurify.jl:39
>   Stacktrace:
>    [1] forward_linear(::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}, ::Layer{ReLU,Float64}) at /Users/changliuliu/.julia/packages/NeuralVerification/1cnod/src/adversarial/reluVal.jl:111
>    [2] forward_layer(::ReluVal, ::Layer{ReLU,Float64}, ::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}) at /Users/changliuliu/.julia/packages/NeuralVerification/1cnod/src/adversarial/reluVal.jl:105
>    [3] forward_network(::ReluVal, ::Network, ::Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}) at /Users/changliuliu/.julia/packages/NeuralVerification/1cnod/src/reachability/utils/reachability.jl:9
>    [4] solve(::ReluVal, ::Problem{Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}},Hyperrectangle{Float64,Array{Float64,1},Array{Float64,1}}}) at /Users/changliuliu/.julia/packages/NeuralVerification/1cnod/src/adversarial/reluVal.jl:42
>

tomerarnon:
What do you run that encounters that? Tests pass on my end with the same versions so I'm guessing it isn't that.

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+    r = push!(input.r, interval_width)
+    return SymbolicIntervalGradient(sym, LΛ, UΛ, r)
+end
+
+function forward_act(input::SymbolicIntervalGradient, layer::Layer{Id})
+    sym = input.sym
+    n_node = size(input.sym.Up, 1)
+    LΛ = push!(input.LΛ, ones(Float64, n_node))
+    UΛ = push!(input.UΛ, ones(Float64, n_node))
+    r = push!(input.r, ones(Float64, n_node))
```
tomerarnon, line 286: `r` should be the width of the interval, right? Is it 1.0 here just because we can ignore `r` in the case of `Id` layers later on, so the value doesn't matter?

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+
+# Why can "interval" be an AbstractPolytope?
+struct SymbolicInterval{F<:AbstractPolytope}
+    Low::Matrix{Float64}
+    Up::Matrix{Float64}
+    interval::F
+end
+
+# Data to be passed during forward_layer
+struct SymbolicIntervalGradient
```
tomerarnon, line 41: Is there a conceptual difference between `SymbolicIntervalGradient` and `SymbolicIntervalMask` from reluval? Do `LΛ/UΛ` represent the same relu mask still? If the only different is `r`, I'm wondering whether a refactoring will help consolidate. Does the original paper do something similar to this, or is their approach different?

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+[https://github.com/tcwangshiqi-columbia/Neurify](https://github.com/tcwangshiqi-columbia/Neurify)
+"""
+
+@with_kw struct Neurify
+    max_iter::Int64     = 100
+    tree_search::Symbol = :DFS # only :DFS/:BFS allowed? If so, we should assert this.
+end
+
+# Why can "interval" be an AbstractPolytope?
+struct SymbolicInterval{F<:AbstractPolytope}
```
tomerarnon, line 34: We spoke about star sets in the previous meeting; is this already the stand-in for a star set, or is the idea that eventually `F<:Star` but the rest remains the same? Since `interval` (which may be misnamed now, as already noted) is always the input set, I'm wondering whether a refactoring here makes sense as well.

changliuliu: This is different from a star set. If we only have Low (or Up) and interval, then it is a star set. F will always be <:AbstractPolytope. I suggest changing the `interval` to `domain` to avoid confusion.

src/adversarial/neurify.jl
changliuliu: Yes, `SymbolicIntervalGradient` records gradients as real numbers which can be anything between 0 and 1 (consider relexed_ReLU), `SymbolicIntervalMask` records binary 0/1 gradients as integers. We can actually merge the two definitions. And we probably don't need to record `r`. We can compute it from other variables.

src/adversarial/neurify.jl
changliuliu: This does not look correct to me. @Wei-TianHao

tomerarnon: Any new ideas here?

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+    n_node, n_input = size(input.sym.Up)
+    output_Low, output_Up = copy(input.sym.Low), copy(input.sym.Up)
+    mask_lower, mask_upper = zeros(Float64, n_node), ones(Float64, n_node)
+    interval_width = zeros(Float64, n_node)
+    for i in 1:n_node
+        # Symbolic linear relaxation
+        # This is different from ReluVal
+
+        up_low, up_up = bounds(input.sym.Up[i, :], input.sym.interval)
+        low_low, low_up = bounds(input.sym.Low[i, :], input.sym.interval)
```
tomerarnon, line 260: @changliuliu @Wei-TianHao 
To bring the slack question here:
We first minimize and then maximize the objective (`v'[x; 1]`, where `v` is e.g. `Low[i, :]`) to get the bounds. I think we can replace this instead with the support function/vector from LazySets. What do you think? If I understand correctly what's going on, it would be something like taking the affine map (stored in Up/Low) of `interval`, and then taking the support function in two opposite directions (not sure which though... each dimension independently?)

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+
+        interval_width[i] = up_up - low_low
```
tomerarnon, line 262: @changliuliu you mentioned we can drop the `r` field being made here since this should be calculable using the other fields where we need it in `get_max_nodewise_influence`. How would we calculate it there?

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,306 @@
...
+
+# Symbolic forward_linear for the first layer
+function forward_linear(solver::Neurify, input::AbstractPolytope, layer::Layer)
+    (W, b) = (layer.weights, layer.bias)
+    sym = SymbolicInterval(hcat(W, b), hcat(W, b), input)
+    LΛ = Vector{Vector{Int64}}(undef, 0)
+    UΛ = Vector{Vector{Int64}}(undef, 0)
+    r = Vector{Vector{Int64}}(undef, 0)
+    return SymbolicIntervalGradient(sym, LΛ, UΛ, r)
+end
```
tomerarnon, line 236: To clarify, this method acts on the input set (and does nothing), and then feeds a `SymbolicIntervalGradient` into `forward_act`. Therefor the first layer seems not to go through a `forward_linear`, step. Is that right?

tomerarnon: @changliuliu @Wei-TianHao

changliuliu: It does go through the first layer, `sym` is computed using the parameters in the first layer.

src/adversarial/neurify.jl
changliuliu: We just need to do the same calculation here. Get the symbolic representation `sym`, then compute `up_up` and `low_low`. Then `r` is the difference between the two.

src/adversarial/neurify.jl
changliuliu: Yes, using LazySets would be a good idea.

src/adversarial/neurify.jl
tomerarnon: I see what's going on now (the same happens in ReluVal as well, and both algorithms are currently correct, too, so there is no question about that). I guess the "most technically correct" thing to do would be to initialize Up and Low as `[I zeros(N)]` and then send them through a normal forward_linear step to turn that into `[W b]`...

changliuliu: Yes, that may be a better way.

src/adversarial/neurify.jl
tomerarnon: Looks like where we need `r` we've actually lost most of the info needed to easily get it (in the backprop in nodewise_influence). What we have there is the symbolic interval at the output, and we need the symbolic layer at each layer `i` to get its radius. That's a bit annoying...

src/adversarial/neurify.jl
```diff
@@ -0,0 +1,297 @@
...
+    # Backpropagation to calculate the node-wise gradient
+    for i in reverse(1:length(nnet.layers))
+        layer = nnet.layers[i]
+        if layer.activation isa ReLU
+            for j in 1:n_nodes(layer)
+                if is_ambiguous_activation(i, j)
+                    # taking `influence = max_gradient * reach.r[i][j]*k` would be
+                    # different from original paper, but can improve the split efficiency.
+                    # where `k = n-i+1`, i.e. counts up from 1 as you go back in layers.
+                    influence = max(abs(LG[j]), abs(UG[j])) * radii[i][j]
```
tomerarnon, line 211: @Wei-TianHao do you mind explaining this comment? I did edit it from the original wording since I changed some of the code around it, but I never understood it fully then either. It would be useful in case we want to adopt that method in the future to know why/how it is more efficient.

Issue #46, opened 2019-02-19
Update LazySets


tomerarnon:
LazySets has made a lot of updates since we pinned NV to a particular version. Some of those changes are breaking, so we can't just upgrade without updating things on our end, but updating also offers performance gains.

mforets:
> but updating also offers performance gains.

LazySets [v1.7.0](https://github.com/JuliaReach/LazySets.jl/releases/tag/v1.7.0) implements a simple, "low-hanging fruit" improvement for the concrete linear map of polyhedra; this could replace and in some cases improve [passing to vrep in this function](https://github.com/sisl/NeuralVerification.jl/blob/master/src/utils/util.jl#L332); see [linear_map docs](https://juliareach.github.io/LazySets.jl/latest/lib/interfaces.html#LazySets.linear_map-Union{Tuple{N},%20Tuple{AbstractArray{N,2},AbstractPolyhedron{N}}}%20where%20N%3C:Real ) in LazySets for all the available options.

Perhaps more interesting here is that the same trick applies if the map is affine, such as for a [linear_transformation(layer::Layer, input::HPolytope)](https://github.com/sisl/NeuralVerification.jl/blob/master/src/utils/util.jl#L313). This idea is explained in Lemma 2 of [Computational Techniques for Hybrid Systems Verification, A. Chutinan and B. H. Krogh, IEEE Transactions on Automatic Control](https://pdfs.semanticscholar.org/0aa8/337a5447c4b3eed9d2fb49d766b318fc9356.pdf).

tomerarnon:
@mforets, indeed, that concrete linear map in h-rep is the primary improvement I was referring to. Also, thanks for the link to the explanation!

I recall you proposed adding an `AffineMap`/`affine_map` functionality to LazySets. Is this something that is still planned? We are thinking of renaming `linear_transformation`, which is verbose (and not technically correct), to `affine_map` as per LazySets convention. I don't think this would clash with your naming, even if you did add the function, since then we could simply use yours as a drop-in replacement, but still I am curious.

tomerarnon:
@mforets, on this same topic, is there currently a way to perform a concrete translation of a set (an HPolytope in particular). I think one is proposed to go along with `AffineMap`, but just wanted to make sure none existed already (couldn't find any mention, so it's a safe bet...).
[...]

Issue #83, opened 2020-02-10
Inconsistent bounds in Planet


mforets:
If you try run the test script for `identity_network.jl` checking the construction of hyperrectangles in the function `tighten_bounds` and showing the data in case of failure, like this,

```julia
    new_bounds = Vector{Hyperrectangle}(undef, length(neurons))
    for i in 1:length(neurons)
        try
            new_bounds[i] = Hyperrectangle(low = lower[i], high = upper[i])
        catch
            @show i
            @show lower[i]
            @show upper[i]
        end
    end
    return (OPTIMAL, new_bounds)
```
The result is
```julia
i = 1
lower[i] = [0.9]
upper[i] = [-0.9]
i = 2
lower[i] = [2.3999999999999986, 2.4000000000000004]
upper[i] = [0.6000000000000001, 0.6]

i = 3
lower[i] = [12.100000000000001, 12.100000000000001]
upper[i] = [4.8999999999999995, 4.8999999999999995]
```
Hence LazySets refuses to create an hyperrectangle given these bounds (for reference, we added this consistency check in the inner constructor in [v1.16.0](https://github.com/JuliaReach/LazySets.jl/releases/tag/v1.16.0)).

Would it be reasonable to discard the data for which the construction criterion doesn't hold, at the risk of returning a number `< length(neurons)` of neurons? (If not, what is the meaning of such sets anyway, an empty set?)

tomerarnon:
Hi @mforets, thanks for opening this issue as a follow up to you pr!
@changliuliu and I have taken and look at this and it turned out that any bounds where `lower > upper` indicate the problem is infeasible. However, patching this by making `lower <= upper` a strict requirement (may have) revealed another issue that went unnoticed, so I can't yet push it as a fix. We're looking into that currently.

mforets:
Alright, thanks for taking a look!

mforets:
Hi, do you happen to have any updates on this issue? Here is a small one: [LazySets v1.31.0](https://github.com/JuliaReach/LazySets.jl/releases/tag/v1.31.0) adds an option to the hyperrectangle constructor, so `Hyperrectangle(low = lower[i], high = upper[i], check_bounds=false)` can be used to ignore the bounds check, in case you still consider this as a temporary workaround for this algorithm. This change is less intrusive than globally deactivating all the assertions.

changliuliu:
[...]

PR #84, opened 2020-02-10
#46 - Update LazySets

mforets:
This PR has two purposes:

- Bump the package to use the latest released version of LazySets. The tests in NeuralVerification now pass, but see (*).
- Cleanup of the concrete affine map functions, using those provided from LazySets directly (see [this thread](https://github.com/sisl/NeuralVerification.jl/issues/46)).

(*) To make the tests pass, I have deactivated assertions (I don't recommend to keep this behavior.. but for the purpose of debugging it's useful to have!), see the new lines in `runtests.jl`. With assertions activated, I get the error reported here https://github.com/sisl/NeuralVerification.jl/issues/83.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,9 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 
+[compat]
+LazySets = "1.27.0"
+
 [extras]
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/Project.toml
+++ b/Project.toml
@@ -17,7 +17,7 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 
 [compat]
-LazySets = "1.27.0"
+LazySets = "1.29.0"
 
 [extras]
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
--- a/src/NeuralVerification.jl
+++ b/src/NeuralVerification.jl
@@ -11,7 +11,7 @@ using LinearAlgebra
 using Parameters
 using Interpolations # only for PiecewiseLinear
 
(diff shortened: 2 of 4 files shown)


tomerarnon:
Hi @mforets, thanks for putting this update together! Also, _super cool_ that you guys implemented `LazySets.deactivate_assertions`! It's a great idea for debugging, as you say.

As for the error you saw in Planet, I'm not sure what it means for those bounds to be inconsistent. I.e. whether it indeed is an empty set or whether `lower/upper` always represent a valid interval (and just need to be sorted). I've pinged @changliuliu about it, and hopefully we can figure out the correct way forward. I'll update on #83 with the answer once it's known.

Looks like the travis build isn't succeeding here because the Manifest file isn't updated (that's my guess at least). Since this is on your fork, would you mind `up`ing LazySets in this pr (or `free`ing it, if that also works)?

mforets:
> Bump the package to use the latest released version of LazySets.

That was too fast, sorry -- the required changes have not been tagged yet, but we plan to do so within 1 week. You can try this branch on top of `LazySets#mforets/nverif`, that works. 

> Since this is on your fork, would you mind uping LazySets in this pr (or freeing it, if that also works)?

Not sure that I follow, do you suggest to change the Manifest file? In that case, we will most probably need to change several entries in that file to resolve a set of compatible versions, and that takes time.

It is not customary to manage the dependencies with the Manifest; instead, we update the compability section of the `Project.toml` file. Have you considered saving the manifest in another place? (eg. a `repeatability` folder) Also, the update process is simplified with existing tools (https://github.com/bcbi/CompatHelper.jl).

tomerarnon:
> It is not customary to manage the dependencies with the Manifest; instead, we update the compability section of the Project.toml file. 

I have precisely been using the Manifest in this way. I will look into swapping it out in favor of a `compat` section in the Project file. Thanks for the tip, and for the link to CompatHelper (seems pretty convenient!)

Project.toml
```diff
@@ -16,6 +16,9 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 
+[compat]
+LazySets = "1.27.0"
```
mforets:
```suggestion
LazySets = "1.29.0"
```

mforets:
Closing, see [this comment](https://github.com/sisl/NeuralVerification.jl/pull/88#issuecomment-646000127).

PR #88, opened 2020-02-24
Update Project.toml

mforets:
This PR adds compatibility bounds to all dependencies. In doubt about the version number, I allowed more than one minor version.


tomerarnon:
Thanks @mforets for opening this pull request! Following your suggestion in #46, and noting that the tests on this pr fail due to a conflict between the manifest and the new `compat` section, I have removed Manifest.toml (and in the process learned how to navigate other people's forks in git 😁). I'll give the tests another chance to run, and if all goes smoothly, I am open to merging it.

mforets:
Sounds good!

>  We are still looking into why tighten_bounds in Planet sometimes returns "inverted" bounds

Just a note: from the LazySets side, we could make the check optional, eg. by adding a flag in the `Hyperrectangle` constructor, `check_bounds`, that can be optionally turned off. (I think it's a good idea to have that as an option anyways).

tomerarnon:
> Just a note: from the LazySets side, we could make the check optional, eg. by adding a flag in the Hyperrectangle constructor, check_bounds, that can be optionally turned off. (I think it's a good idea to have that as an option anyways).

Interesting! In that case, what would be the representation (and behavior) of `radius`? Would it be an absolute value or the raw values? By extension also would `high/low` return the lower and upper bound respectively (or work "correctly" and return the reverse i.e. `high` would return the higher value regardless of the construction details)?

mforets:
I think it should be "undefined" (can crash, use at your own risk, etc.). Compare to `IntervalArithmetic.interval` which checks that the interval is well formed, and `IntervalArithmetic.Interval` which doesn't peform the check.

```julia
julia> IntervalArithmetic.interval(2, 1)
ERROR: ArgumentError: `[2, 1]` is not a valid interval. Need `a ≤ b` to construct `interval(a, b)`.
Stacktrace:
 [1] interval(::Int64, ::Int64) at /home/mforets/.julia/packages/IntervalArithmetic/lyDPa/src/intervals/intervals.jl:111
 [2] top-level scope at REPL[5]:1

julia> IntervalArithmetic.Interval(2, 1)
[2, 1]

julia> diam(ans)
-1.0
```
This option would be more for debugging purposes, or speedup the creation of a very large number of rectangles (which you know are well defined).

mforets:
I moved this discussion to a new issue in LazySets (https://github.com/JuliaReach/LazySets.jl/issues/2015).

mforets:
Bump?

mforets:
Closing as this is now outdated -- see [our fork](https://github.com/JuliaReach/NeuralVerification.jl) for the updated dependencies file, and feel free to ping me if I can help with a PR again to this repo!

PR #93, opened 2020-03-17
Avoid model recreation at every call of reluplex_step

castrong:
Currently, the model is recreated whenever a ReLU is fixed. This changes it so that (1) the desired constraints are added to the model to fix the ReLU to be active or inactive, (2) the recursive call is made, then (3) the constraints are removed from the model. This is now possible b/c of added functionality to the JuMP optimization library.

### DIFF
--- a/src/satisfiability/reluplex.jl
+++ b/src/satisfiability/reluplex.jl
@@ -119,14 +119,13 @@ function reluplex_step(solver::Reluplex,
 
         for repair! in (type_one_repair!, type_two_repair!)
             # Add the constraints associated with the ReLU being fixed
-            con_one, con_two = repair!(model, ẑ[i][j], z[i][j])
+            new_constraints = repair!(model, ẑ[i][j], z[i][j])
 
             # Recurse with the ReLU i, j fixed to active or inactive
             result = reluplex_step(solver, problem, model, ẑ, z, relu_status)
 
             # Reset the model when we're done with this ReLU
-            delete(model, con_one)
-            delete(model, con_two)
+            delete.(model, new_constraints)
 
             result.status == :violated && return result
         end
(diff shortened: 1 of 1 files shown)


src/satisfiability/reluplex.jl
```diff
@@ -131,17 +133,24 @@ function reluplex_step(solver::Reluplex,
...
-            new_m  = Model(solver)
-            bs, fs = encode(solver, new_m, problem)
-            enforce_repairs!(new_m, bs, fs, relu_status)
-
-            result = reluplex_step(solver, problem, new_m, bs, fs, relu_status)
-
-            # Reset the relu when we're done with it.
+            if (repair_type == 1)
+                con_one, con_two = type_one_repair!(model, ẑ[i][j], z[i][j])
+                print("Con one, two: ", con_one, con_two)
```
mykelk: Looks like debug code was left in?

castrong: yeah sorry - just removed them

castrong:
@tomerarnon I was considering keeping relu_status around for debugging purpose but it would also be cleaner without. I'll try to implement some of the changes you suggested - I'm not sure I get what you mean by the entire function goes away? Could you elaborate on that a bit more?

Thanks!
Chris

tomerarnon:
It was a typo! I mean the entire `enforce_repairs` function goes away.

tomerarnon:
By the way, it seems the tests are failing currently because `con_one`/`con_two` can't be reassigned using that syntax. 
The error says:
`
 An object of name con_one is already attached to this model. If this is intended, consider using the anonymous construction syntax, e.g., x = @variable(model, [1:N], ...) where the name of the object does not appear inside the macro.
`

castrong:
Ok I tried changing that and refactored the for loop to be cleaner as suggested :) - I'll see if it breaks any tests now

tomerarnon:
Looks like that fixed it!

One final stylistic suggestion before merging: how about instead of `con_one`/`two`, having something like
```julia
new_constraints = repair!(...)
...
delete.(model, new_constraints)
```

castrong:
sounds great - just pushed that - also should I be worried about the coverage/coveralls check?

tomerarnon:
> just pushed that

Awesome. We already know it works, but I'll wait for the tests out of paranoia anyway.

>should I be worried about the coverage/coveralls check

Nope. <1% in either direction isn't meaningful anyway.

PR #95, opened 2020-04-20
Fixed neurify, runtests passed.

Wei-TianHao:
Fixed the Neurify's implementation based on Changliu's code. ReluVal is also modified to share some common functions with Neurify. Added a new helper function in util.jl.

Passed runtests.jl and runtests2.jl (If commented the Reluplex part).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,13 @@ language: julia
 os:
   - linux
   - osx
+  - windows
 
 julia:
   - 1.0
   - 1.1
+  - 1.2
+  - 1.3
   - nightly
 
 notifications:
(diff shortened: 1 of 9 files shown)


tomerarnon:
Changed the base branch of this PR to be the neurify branch as per #79 .
This introduced a few merge conflicts, so let me know if this is not the intention and we can easily reverse it!

Wei-TianHao:
I resolved the conflicts online, it should be fine now. Please let me know if it doesn't work. Thanks

tomerarnon:
Looks like the tests are failing (the badge isn't updating, but if you click on details you can see). It seems something may have changed in the tests themselves. Do you mind looking into what happened?

PR #99, opened 2020-06-28
Fix Duality Fix #16

### LINKED ISSUE #16: JuMP.solve in duality
Duality is getting stuck on `solve`.

changliuliu:
Update the activation_value function in duality. Now duality can properly return solutions and pass the runtests.jl except it returns violated for the hold case in identity_networks.jl.

Since Duality is not a complete method, returning violated for hold cases should be ok. Or we can force it to always return unknown as in other dual optimization approaches: certify and convdual.


src/optimization/utils/variables.jl
```diff
@@ -22,13 +22,13 @@ function init_variables(model::Model, layers::Vector{Layer}; binary = false, inc
...
     for (i, n) in enumerate(all_layers_n)
-        vars[i] = @variable(model, [1:n], binary = binary, base_name = "z$i")
+        vars[i] = @variable(model, [1:n], binary = binary, base_name = string(name, "$i"))
     end
     return vars
 end
 
 function symbolic_max(m::Model, a, b)
-    aux = @variable(m)
+    aux = @variable(m, base_name = "aux_max")
```
tomerarnon, line 31: Unfortunately, we can't do this `base_name = "aux_max"` business so easily here. `symbolic_max` (and the others) will be called more than once per model and a name shouldn't ("can't") be shared between multiple variables. If we really want these to be named we can come up with a scheme to name them sequentially, but it might be a slight hassle.

src/optimization/duality.jl
```diff
@@ -75,9 +75,15 @@ function activation_value(layer::Layer,
     b_hat = approximate_affine_map(layer, bound)
     l_hat, u_hat = low(b_hat), high(b_hat)
     l, u = layer.activation(l_hat), layer.activation(u_hat)
-
-    o += sum(symbolic_max.(μᵢ .* l_hat, μᵢ .* u_hat))
-    o += sum(symbolic_max.(λᵢ .* l,     λᵢ .* u))
+    for j in 1:length(l)
+        if layer.activation == ReLU()
+            s_max = symbolic_max(μᵢ[j] * l_hat[j] - λᵢ[j] * l[j], μᵢ[j] * u_hat[j] - λᵢ[j] * u[j])
+            l_hat[j] * u_hat[j] >= 0 || @constraint(μᵢ[j].model, s_max >= 0.0)
```
tomerarnon, line 81: Just want to confirm that the logic is `l*mu ≥ 0 OR constraint(s ≥ 0)`

tomerarnon: Lets also clean up these lines. e.g. 
```julia
if layer.activation == ReLU()
    s_max = @. symbolic_max(μᵢ*l̂ - λᵢ*l, μᵢ*û - λᵢ*u)
    same_sign_inds = findall(l̂ .* û .>= 0)
    @constraint(μᵢ[same_sign_inds].model, s_max[same_sign_inds] .>= 0.0)
else
    s_max = @. symbolic_max(μᵢ * l̂, μᵢ * û) + symbolic_max(-λᵢ * l, -λᵢ * u)
end
```
and using the updated notation from the paper (`g(l)`, `g(u)`,`g(0)`)

castrong: it seems like the original 

`l_hat[j] * u_hat[j] >= 0 || @constraint(μᵢ[j].model, s_max >= 0.0)`

and the new

`same_sign_inds = findall(l̂ .* û .>= 0)`
    `@constraint(μᵢ[same_sign_inds].model, s_max[same_sign_inds] .>= 0.0)`

Would add constraints in opposite cases. The first would add a constraint when l_hat * u_hat <= 0 is false, while the second would add a constraint when l_hat * u_hat >= 0 is true. I may have flipped something though (I'm still getting used to thinking through the || and && lazy execution).

changliuliu: Nice catch

PR #100, opened 2020-06-30
Add a write_nnet function to utils.jl

castrong:
Added a function write_nnet to utils.jl which writes out a network to a .nnet file. Since the Network object doesn't include upper and lower bounds on the input, we write these to the .nnet file as the maximum and minimum values that a Float16 can take on. 

Also added a file write_nnet_test.jl which provides a single test of the write_nnet function.


src/utils/util.jl
```diff
@@ -47,6 +47,74 @@ function read_layer(output_dim::Int64, f::IOStream, act = ReLU())
...
+    write_nnet(filename, model[; header_text])
+
+Write `network` to \$filename.nnet.
+Note: Does not perform safety checks on inputs, so use with caution.
+
+Based on python code at https://github.com/sisl/NNet/blob/master/utils/writeNNet.py
+and follows .nnet format given here: https://github.com/sisl/NNet.
+"""
+function write_nnet(outfile, network; header_text="Default header text.\nShould replace with the real deal.")
+    name, ext = splitext(outfile, ".")
```
tomerarnon, line 108: It seems this is a mistake. The line should be simply `splitext(outfile)`

tomerarnon:
Looks good to me. Ready to merge it you are.

castrong:
Yeah I'm good to merge it! I don't think I have write access to the repository in order to merge it myself.

PR #101, opened 2020-06-30
Add a function to generate random networks

castrong:
Added a file testing_utils.jl with a function make_random_network which generates a random network with a given size and ranges on its weights and bias. It defaults to use a MarsenneTwister random number generator seeded with 0. I'm not sure if creating the new file testing_utils.jl is the best way to organize things or not.


tomerarnon:
The test failure on linux doesn't seem to have anything to do with this change (failed to install SpecialFunctions.jl and everything else is downstream of that). Merging since that's "not our problem" 🤷‍♂️

PR #102, opened 2020-07-01
Fix neurify

Wei-TianHao:
Add a new run test for Neurify (runtests3). Covering the splitting code. This is a simple R2 to R2 3-layer network.
Fixed a few bugs in node_wise_gradient (used to compute the backward gradient for Neurify).


Wei-TianHao:
Oops, I should merge this to the previous "Neurify" branch pull request.

Wei-TianHao:
It seems I have no way to do it. Sorry for the trouble. I will pull requests from the "Neurify" branch next time.

tomerarnon:
Seems like much of neurify still isn't being hit by the new test. Any ideas why or how to make it happen?

Wei-TianHao:
runtests3.jl seems not executed. Do you have any guess that can lead to this?
runtests3.jl covers most of the code of Neurify on my computer.

And could you please look into the building fail? It shows `Error: Error building CodecZlib`. Which I think may not be caused by my code.

Wei-TianHao:
Oh I see, I should include this test in the runtests.jl. I though all the test files will be executed automatically.

tomerarnon:
> And could you please look into the building fail? It shows `Error: Error building CodecZlib`. Which I think may not be caused by my code.

Yeah, that's not your fault. We can ignore it.

Wei-TianHao:
The splitting test and MNIST test work fine now.

test/splitting.jl
```diff
@@ -0,0 +1,51 @@
...
+    A = [1. 0.; 0. 1.; -1. 0.; 0. -1.;]
+    b = [1.1, 1.1, 0.1, 0.1]
+    X = Hyperrectangle(low=[-1,-1], high=[1,1])
+    # The true output set is a union of two segments : (0,0)->(0,1) U (0,0)->(1,0)
+    Y = HPolytope(A, b) # Output constraints. Hyperrectangle: -0.1, -0.1, 1.1, 1.1
+    problem = Problem(net, X, Y)
+
+    @testset "Neurify" begin
+        solver = Neurify()
+        @test @no_error result = solve(solver, problem)
```
tomerarnon, line 39: Why is it testing no_error here? If the issue is that unknown can come up, we should allow for it explicitly like in the other tests.

test/mnist_1000.jl
```diff
@@ -0,0 +1,82 @@
+using LazySets, Test, LinearAlgebra, GLPKMathProgInterface
+using NeuralVerification
+import NeuralVerification: ReLU, Id
+using JLD2
```
tomerarnon, line 4: Do we need JLD2? It didn't look to me like anything was using it. If not, I'd rather remove the dependency.

src/adversarial/neurify.jl
```diff
@@ -224,6 +272,7 @@ function upper_bound(map::Vector{Float64}, input::HPolytope)
     model =Model(with_optimizer(GLPK.Optimizer))
     @variable(model, x[1:m])
     @constraint(model, [i in 1:n], input.constraints[i].a' * x <= input.constraints[i].b)
+    x = model[:x]
```
tomerarnon, line 275: Is this just for clarity, since the binding for x should already exist after `@variable`?

src/adversarial/neurify.jl
```diff
@@ -26,29 +26,45 @@ Sound but not complete.
 """
 
 @with_kw struct Neurify
-    max_iter::Int64     = 1000
+    max_iter::Int64     = 10
     tree_search::Symbol = :DFS # only :DFS/:BFS allowed? If so, we should assert this.
+    model = Nothing()
+    optimizer = GLPK.Optimizer
+    splitted = Set() # To prevent infinity loop. Not the best solution, need to be modified in the future.
```
tomerarnon, line 33: Lets go with `split` or `splits` or `split_intervals` or something else depending on the spirit of how you meant it, since `splitted` isn't a word

src/adversarial/neurify.jl
```diff
@@ -126,27 +162,39 @@ function constraint_refinement(solver::Neurify, nnet::Network, reach::SymbolicIn
...
     max_tuple = (0, 0, 0.0)
     for (k, layer) in enumerate(reverse(nnet.layers))
         i = n_length - k + 1
-        for j in size(layer.bias)
-            if LΛ[i][j] ∈ (0.0, 1.0) && UΛ[i][j] ∈ (0.0, 1.0)
-                max_gradient = max(abs(LG[j]), abs(UG[j]))
-                if max_gradient > max_tuple[3]
-                    max_tuple = (i, j, max_gradient)
+        if layer.activation != Id() 
+            # Not sure whether this is right, but by experiments, this caused infinity loop.
```
tomerarnon, line 174: @changliuliu do you know the answer here? Also potentially with regards to other changes affecting correctness / logic flow.

src/adversarial/neurify.jl
```diff
@@ -26,29 +26,45 @@ Sound but not complete.
...
+    model = Nothing()
+    optimizer = GLPK.Optimizer
+    splitted = Set() # To prevent infinity loop. Not the best solution, need to be modified in the future.
 end
 
 struct SymbolicInterval{F<:AbstractPolytope}
     Low::Matrix{Float64}
     Up::Matrix{Float64}
     interval::F
+    model
```
tomerarnon, line 40: I'm not seeing this new field used anywhere. Do we need it?

src/adversarial/neurify.jl
```diff
@@ -26,29 +26,45 @@ Sound but not complete.
...
 end
 
-SymbolicInterval(x::Matrix{Float64}, y::Matrix{Float64}) = SymbolicInterval(x, y, Hyperrectangle([0],[0]))
+SymbolicInterval(x::Matrix{Float64}, y::Matrix{Float64}) = SymbolicInterval(x, y, Hyperrectangle([0],[0]), Nothing())
```
tomerarnon, line 43: Just fyi, `nothing` is the commonly used alias for `Nothing()` (I don't mind changing these myself before the merge though, just letting you know.)

src/adversarial/neurify.jl
```diff
@@ -126,27 +162,39 @@ function constraint_refinement(solver::Neurify, nnet::Network, reach::SymbolicIn
...
                 end
             end
         end
         i >= 1 || break
-        LG_hat = max.(LG, 0.0) * Diagonal(LΛ[i]) + min.(LG, 0.0) * Diagonal(UΛ[i])
-        UG_hat = min.(UG, 0.0) * Diagonal(LΛ[i]) + max.(UG, 0.0) * Diagonal(UΛ[i])
-        LG, UG = interval_map_right(layer.weights, LG_hat, UG_hat)
+        LG_hat = Diagonal(LΛ[i]) * max.(LG, 0.0) + Diagonal(UΛ[i]) * min.(LG, 0.0)
+        UG_hat = Diagonal(LΛ[i]) * min.(UG, 0.0) + Diagonal(UΛ[i]) * max.(UG, 0.0)
+        LG, UG = interval_map(copy(transpose(layer.weights)), LG_hat, UG_hat)
```
tomerarnon, line 195: I'm guessing you added the copy to avoid an error associated with `Adjoint{...}`? If so, the right thing to do is actually to change the type annotation in the definition of interval_map (see below) to `AbstractMatrix`
https://github.com/sisl/NeuralVerification.jl/blob/b99f427a77538f0863899b9c2f5d2ff68856759f/src/utils/util.jl#L256

src/adversarial/neurify.jl
```diff
@@ -126,27 +162,39 @@ function constraint_refinement(solver::Neurify, nnet::Network, reach::SymbolicIn
...
     end
+    push!(solver.splitted, (max_tuple[1], max_tuple[2]))
```
tomerarnon, line 197: @changliuliu we're discarding gradient information here. Would there be any value in keeping it?

src/adversarial/neurify.jl
```diff
@@ -70,47 +86,67 @@ function check_inclusion(reach::SymbolicInterval{HPolytope{N}}, output::Abstract
...
-        end
-        if objective_value(model) > output.constraints[i].b
             return BasicResult(:unknown)
         end
         
     end
     return BasicResult(:holds)
 end
 
+function printconvex(shape)
```
tomerarnon, line 137: Just for debugging, right? I don't mind keeping it around, but if you don't think you'll need it again, we can just go ahead and delete it.

test/splitting.jl
Wei-TianHao: I thought `no_error` is a must be to convert a running status to a `true` or `false` bool. 
Looks like I'm wrong. I will remove it.

test/mnist_1000.jl
Wei-TianHao: I stored 1000 mnist images in the file `mnist_1000.jld2`.
The file is loaded in `mnist_1000.jl` at line 19.

src/adversarial/neurify.jl
Wei-TianHao: It's just a redundancy, I will remove it.

src/adversarial/neurify.jl
Wei-TianHao: `splitted` is a set to prevent splitting a ReLU node more than one time. What do you think is a proper name?

src/adversarial/neurify.jl
Wei-TianHao: This comment is out-dated. What I was not sure is whether this could prevent an infinity loop. But later, I defined the `splitted` set, which can for sure prevent an infinity loop.
The logic here should be right, the intuition is only ReLU nodes should be split to reduce over-approximation.
I will change the comment.

src/adversarial/neurify.jl
Wei-TianHao: Not anymore. Will remove.

src/adversarial/neurify.jl
Wei-TianHao: Will remove. Since `model` in `SymbolicInterval` will be gone.

src/adversarial/neurify.jl
Wei-TianHao: Will do.

src/adversarial/neurify.jl
Wei-TianHao: Will remove.

src/adversarial/neurify.jl
Wei-TianHao: I change it to `splits`

test/splitting.jl
tomerarnon: It simply tests that a line didn't produce an error. We used it once or twice on tests that we didn't have a confirmed answer for yet but wanted to see that the code at least ran. We're phasing it out where possible.

src/adversarial/neurify.jl
Wei-TianHao: It's a good idea to also push gradient into this set. It allows a node to split multiple times without causing a potential infinity loop. 
Will do.

test/mnist_1000.jl
tomerarnon: Ah ok, and looks like you set the requirement as test-only, so it's fine by me.

PR #106, opened 2020-07-07
Nodes

tomerarnon:
Noticed in #105 that `Node` really isn't required for maxsens. 
It also should be quite possible to write it using our own `approximate_affine_map` or LazySet's `overapproximate` (probably). For example, we can write it as 
```julia 
function forward_layer(solver::MaxSens, L::Layer, input::Hyperrectangle)
    output = approximate_affine_map(L,  input)
    β = L.activation.(center(output))
    βmax = L.activation.(high(output)
    βmin = L.activation.(low(deviation))
    #etc.
```

### DIFF
--- a/src/reachability/maxSens.jl
+++ b/src/reachability/maxSens.jl
@@ -41,27 +41,25 @@ end
 # This function is called by forward_network
 function forward_layer(solver::MaxSens, L::Layer, input::Hyperrectangle)
     (W, b, act) = (L.weights, L.bias, L.activation)
-    center = zeros(size(W, 1))
-    gamma  = zeros(size(W, 1))
-    for j in 1:size(W, 1)
-        node = Node(W[j,:], b[j], act)
-        center[j], gamma[j] = forward_node(solver, node, input)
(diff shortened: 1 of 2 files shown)


src/reachability/maxSens.jl
```diff
@@ -40,53 +40,48 @@ end
...
-    (W, b, act) = (L.weights, L.bias, L.activation)
-    center = zeros(size(W, 1))
-    gamma  = zeros(size(W, 1))
-    for j in 1:size(W, 1)
-        node = Node(W[j,:], b[j], act)
-        center[j], gamma[j] = forward_node(solver, node, input)
-    end
-    return Hyperrectangle(center, gamma)
-end
+    output = approximate_affine_map(L,  input)
```
tomerarnon, line 43: `output = overapproximate(AffineMap(L.weights, input, L.bias))` is the other option. Any preference on which is clearer for the paper @changliuliu?

changliuliu: I like `approximate_affine_map`

PR #115, opened 2020-07-09
Testing setup

castrong:
Create a system for testing correctness of the different algorithms. This includes code to (i) write problems to file, (ii) generate random networks, input sets, and output sets and combine these into a problem, and (iii) run tests with these problems, comparing the results of different solvers for consistency. This includes a small, medium, and large test set of randomly generated problems. At the moment inconsistencies found so far have been added as issues.


castrong:
Replaced with #175 which cleared some of the big file adds and removes.

PR #116, opened 2020-07-12
tests pass with latest versions

tomerarnon:
This pr:
- brings back the `Solver` abstract type to resolve type piracy with `JuMP`. All solvers are subtypes of it (and Neurify, pending merge, will need to subtype it as well)
- deletes Manifest.toml
- deletes unused `symbolic_<>` functions
- changes how the `model` is extracted from variables/expressions in `variables.jl`


tomerarnon:
This PR also, quite unfortunately, breaks support for julia 1.0-1.1 due to unsatisfiable constraints for a second-level dependency (and also breaks support for windows). @mforets, I was hoping you could lend experience in how to resolve these issues. To deal with 1.0+, we can tag a new release prior to this PR and migrate bug fixes to it separately (and potentially drop support later on), but since I noted LazySets supports 1.0-1.4, I'm wondering if you have a workflow for dealing with a dependency's dependencies when they fall out of line and cause these sorts of problems.

As for windows, we were tracking the master branch of PicoSat.jl previously, I think by virtue of the Manifest. I have reached out to the maintainer to tag a new release with the changes that are on master, but do you know of a way to track the master branch of a package without the use of the Manifest.toml file?

mforets:
> Edit: There is one other matter, which appears in the successful builds, which is a repeated warning from Polyhedra.

I think that's fixed in this patch: https://github.com/JuliaReach/LazySets.jl/pull/2195, but it hasn't yet been merged on our end.

mykelk:
I'm okay with focusing on only supporting more recent versions of Julia.

mforets:
> Edit: There is one other matter, which appears in the successful builds, which is a repeated warning from Polyhedra.

@tomerarnon  LazySets v1.27.0 has been tagged, i wonder if that solves the issue?

>  if you have a workflow for dealing with a dependency's dependencies when they fall out of line and cause these sorts of problems.

if you need to run different code for different Julia versions, the usual way to do this is with a branch, that you can annotate with the `@static` flag, for example

```julia
@static if VERSION < v"1.2"
    LinearAlgebra.rank(M::SparseMatrixCSC) = rank(qr(M)) # rank of sparse matrix (see JuliaLang #30415)
end
```

>  but do you know of a way to track the master branch of a package without the use of the Manifest.toml file?

do you mean in the travis script? you can add custom package specs to the build. [here](https://github.com/JuliaReach/NeuralNetworkAnalysis.jl/blob/master/.travis.yml#L35) is an example.

tomerarnon:
> @tomerarnon LazySets v1.27.0 has been tagged, i wonder if that solves the issue?

Next time the tests run we'll see.

> if you need to run different code for different Julia versions, the usual way to do this is with a branch, that you can annotate with the @static flag, for example

More along the lines of different dependencies for different versions. I.e. a <v1.2 set of compatible dependency and a 1.2+ version. I can use a build.jl with those sort of annotations to build up the dependencies manually, but I think that's rather messy. We are already resigned to drop support for 1.0-1.1 though (if you know another solution though, please do share!).

>do you mean in the travis script? you can add custom package specs to the build. here is an example.

I meant as the "compatible" version for the repo. The latest release of that package doesn't support Windows, while the master branch does. I added a build.jl and am using a PackageSpec to get the master version, but for some reason that is only working for two of the Windows builds. Not sure how to explain that yet.

mforets:
> if you know another solution though, please do share!

alright :+1:  no, i don't have any particular suggestion to share, sorry. but folks on [discourse.julialang.org](https://discourse.julialang.org/) may help.

Issue #107, opened 2020-07-09
Inconsistent Results  between Reluplex and NSVerify


castrong:
I'm seeing an inconsistent results between Reluplex and NSVerify, both complete solvers, on the same problem. I believe that NSVerify is correct since the counter-example it returns back is a valid counterexample (shown in the code below). Code is below, with the .nnet file given at the bottom. To replicate this issue copy the network into a file and set the filename in the first line to point to that file.

```
network = NeuralVerification.read_nnet("./test_rand/networks/rand_[1,3,1]_2.nnet")
input_set = NeuralVerification.Hyperrectangle([0.3791619134008495], [0.10069447579678581])
output_set = NeuralVerification.PolytopeComplement(NeuralVerification.HPolytope([NeuralVerification.HalfSpace([1.0], 0.535032787547191), NeuralVerification.HalfSpace([-1.0], 0.8711752040844563)]))
problem = NeuralVerification.Problem(network, input_set, output_set)

# Solve on Reluplex and NSVerify
println(NeuralVerification.solve(NeuralVerification.Reluplex(), problem))
println(NeuralVerification.solve(NeuralVerification.NSVerify(), problem))

# Test the counterexample
println("Counter example is outside set: ", NeuralVerification.compute_output(network, [0.2784674376040637]) ∈ output_set.P) # It's in 

```

CounterExampleResult(:holds, Float64[])
CounterExampleResult(:violated, [0.2784674376040637])
Counter example is outside set: true

.nnet file pasted in:

//// Neural Network File Format by Kyle Julian, Stanford 2016
2, 1, 1, 3,
1, 3, 1,
This line extraneous
-6.55e4,
6.55e4,
0.0,0.0,
1.0,1.0,
-0.2092992410225305,
0.1527785697079942,
0.6952328858495411,
-0.12896309655879667,
0.13005202010414774,
-0.34826001136290063,
0.004074797378236905, 0.8952987002088535, -0.07894098008654638,
-0.8416186860567856,

AvrahamRaviv:
Hi
I'd tackled the same problem.
Since NSVerify returns a counterexample, it should be a bug in Reluplex's implementation.
BTW I've cases that Reluplex returns a counterexample but NSVerify returns hold.
If you find an answer for this Inconsistent - let me know, please.

castrong:
Will do - do you still have access to those examples where Reluplex returns a counterexample but NSVerify returns hold? If so would you be able to share such an example?

Thanks!

[...]

Issue #90, opened 2020-02-26
comparison tolerance in reluplex [and other algorithms]


tomerarnon:
@castrong brought this issue up, which may be impacting his results.

See e.g., the following two lines in reluplex, and in particular the comparisons with 0

```julia
type_one_broken(ẑᵢⱼ, zᵢⱼ) = (zᵢⱼ > 0.0)  && (zᵢⱼ != ẑᵢⱼ) 
type_two_broken(ẑᵢⱼ, zᵢⱼ) = (zᵢⱼ == 0.0) && (ẑᵢⱼ > 0.0)
```

Due to the precision of the underlying solvers, sometimes numbers on the order of 1e-10 can be returned, and should probably be considered zeros. We can address this is by introducing a tolerance:
```julia
type_one_broken(ẑᵢⱼ, zᵢⱼ) = (zᵢⱼ > TOL)  && (zᵢⱼ != ẑᵢⱼ)  
type_two_broken(ẑᵢⱼ, zᵢⱼ) = (-TOL < zᵢⱼ < TOL) && (ẑᵢⱼ > 0.0)
```
This is just one example; other optimization solvers also perform similar comparisons. Note that the`-TOL` is maybe not desirable; it may by more principled to just say `z<TOL`, not sure. 

Does this sound like the right fix for this @changliuliu @mykelk, or do we inadvertantly invalidate completeness if we do this? Is there a principled way of picking the tolerance? We can set it to be user-defined in the Reluplex type, but we'd still have to come up with a default.

mykelk:
I'm okay with this. These numerical issues are pretty tricky.

castrong:
I wonder if Julia's ≈ (equivalent to [Base.isapprox](https://docs.julialang.org/en/v1/base/math/#Mathematical-Functions-1) with default arguments) operator could be helpful? It does some sort of approximate equality using tolerances - I don't know if there's an equivalent built in function for comparisons.

mforets:
Since this library uses LazySets, I would suggest that you control tolerance with `LazySets.Tolerance`, documented [here](https://juliareach.github.io/LazySets.jl/dev/lib/comparisons/#LazySets.Tolerance) and the default values are implemented [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Utils/comparisons.jl#L55). In particular you can *globally* control the tolerance for set operations (membership tests, inclusion checks etc) for a given numeric type with the `set_xtol` functions. 

[...]

Issue #82, opened 2020-02-05
Applying layer-bounds in reluplex leads to false :holds


tomerarnon:
-- update --
Actually for this other simple network, `tight = false` doesn't solve the problem:
```julia
l1 = Layer(reshape([-3.0; 1.7], (2, 1)), [1.2, -4.8], ReLU())
l2 = Layer([-0.58 -1.37], [3.94], ReLU())
nnet = Network([l1, l2])
```
However, this is still definitely a bounds issue since completely removing the layer-bounds from reluplex leads to the correct result. 

It isn't clear whether the bound calculation is intrinsically wrong somehow, or that its usage isn't quite correct; still needs to be explored.

-- original issue --

The incorrect bounds affect other solvers that use them as constraints, such as Reluplex. This failure case is courtesy of Chris Strong:
```julia
nnet = Network([Layer(reshape([1.0; 1.0], (2, 1)), [0.5, 0.1], ReLU()), 
                Layer([-4. -1.], [1.0], Id())])

# on the interval [-1, 1], nnet ranges in [1, -6.1]. We test
# y ≤ 0.05 | x ∈ [-1, 1]  which we know is false.
input_set = Hyperrectangle(low = [-1.0], high = [1.0])
output_set = HalfSpace([1.0], 0.05)
problem = Problem(nnet, input_set, output_set)

julia> solve(Reluplex(), problem)
CounterExampleResult(:holds, Float64[])
```
This is obviously incorrect, and is due to the bounds being overly tight (leading to an infeasible optimization problem). Changing the MaxSens calculation to `tight = false` yields the correct result (i.e. `:violated`). See these sections for reference on where this is happening:

The `tight` version is used in `get_bounds`:

https://github.com/sisl/NeuralVerification.jl/blob/b99f427a77538f0863899b9c2f5d2ff68856759f/src/utils/util.jl#L271-L273

For the return statement in question, see line 60:
https://github.com/sisl/NeuralVerification.jl/blob/cab24c5c2203be609c6a7396a59cb5e15a4ce322/src/reachability/maxSens.jl#L53-L64

@changliuliu, before I go any deeper, is there anything above that strikes you as incorrect?

Issue #70, opened 2019-08-13
Inconsistent Results from NSVerify, ReluVal, Reluplex


tclayton0594:
Hi, I am opening this new issue as a follow-up discussion I had with @tomerarnon. It was brought to my attention that ReluVal may not be a complete solver, but to my knowledge NSVerify and Reluplex should be. For the input set, I defined upper and lower bound arrays for my 5-D input and then formed a hyperrectangle polytope. For the output set, I made 2 seperate: 1 for ReluVal and the other for NSVerify and Reluplex since they need a PC output. Since ranges form convex sets, I defined the PC with solely an upper bound of 287.1312. Here is a screenshot of how I setup the parameters for the solvers. 

![Reluplex Reluval Problem Setup](https://user-images.githubusercontent.com/32907889/62949292-ee84ff80-bdb3-11e9-8675-08eda873ddb4.PNG)

The network I am working with has an input size of 5, 3 hidden layers with 30 neurons each, and a single output neuron. Below is a screenshot of the results I got. 

![Reluplex Reluval Problem Results](https://user-images.githubusercontent.com/32907889/62949485-3f94f380-bdb4-11e9-98d1-78ce774bf0aa.PNG)

It is interesting that the counter example returned by Reluval is the midpoint of my input space, but it is unfortunately not a valid counter example. Both Reluplex and NSVerify use the same input/output types and are listed as complete, yet they gave me contradictory results. NSVerify indeed returned a valid counter example point. 

I also went a step further and narrowed my input domain around the violation found by NSVerify and ReluVal was able to find a valid counter example, but Reluplex still returned an incorrect holds. Any thoughts on this/is my formatting of the input and output spaces correct?

tomerarnon:
Thanks for this @tclayton0594, I'm looking into it. Without your network I cannot reproduce this exactly. You can email me it if you'd like to (and indeed can) share it privately. At the moment I am just trying to determine what could have led to this failure path for each of the algorithms to help debug.

[...]

PR #117, opened 2020-07-13
Redo Change for Bounds

castrong:
Redo the change in reluplex to apply the maxsens bounds post activation instead of pre-activation. Addresses issue #107.


tomerarnon:
Also addresses #70 and #82, which we noted quite some time ago were all caused by the same thing.

castrong:
I have also added in a tolerance to account for numerical instability in Reluplex. This is associated with issue #90, and fixes an infinite recursion where the same node is repeatedly fixed in Reluplex. Similar numerical stability changes for other algorithms should be added eventually as well.

Issue #118, opened 2020-07-14
Should NSVerify warn if big-M value is too small?


castrong:
In issue #107  @AvrahamRaviv brought up an example where m = 100 led to a false hold result with NSVerify. With a larger m value the solver worked. Problematic networks are given [here](https://drive.google.com/drive/folders/1uL1ONmoBy5zdgQa_HlSJFc1rvxAZBrhE) and code to replicate the issue is shown below:

```
using NeuralVerification, LazySets, Test, LinearAlgebra, GLPKMathProgInterface
import NeuralVerification: ReLU, Id

mnist1_file = "$(@__DIR__)/networks/mnist_network_new7.nnet"
mnist1 = read_nnet(mnist1_file, last_layer_activation = Id())

[...]

PR #120, opened 2020-07-18
automate calculating M

tomerarnon:
addresses #118


tomerarnon:
The build fails because of an issue getting the latest version of PicoSAT to work with windows.

PR #125, opened 2020-07-23
Add Ai2z solver

SebastianGuadalupe:
It still remains to integrate this with Ai2, as suggested in this comment: https://github.com/sisl/NeuralVerification.jl/issues/87#issuecomment-662742086
I will open an issue to do this in the future

### DIFF
--- a/src/reachability/utils/reachability.jl
+++ b/src/reachability/utils/reachability.jl
@@ -3,7 +3,7 @@
 # Performs layer-by-layer propagation
 # It is called by all solvers under reachability
 # TODO: also called by ReluVal and FastLin, so move to general utils (or to network.jl)
-function forward_network(solver, nnet::Network, input::LazySet)
+function forward_network(solver, nnet::Network, input::AbstractPolytope)
     reach = input
     for layer in nnet.layers
         reach = forward_layer(solver, layer, reach)
--- a/src/reachability/utils/reachability.jl
+++ b/src/reachability/utils/reachability.jl
@@ -3,7 +3,7 @@
 # Performs layer-by-layer propagation
 # It is called by all solvers under reachability
 # TODO: also called by ReluVal and FastLin, so move to general utils (or to network.jl)
-function forward_network(solver, nnet::Network, input::AbstractPolytope)
+function forward_network(solver, nnet::Network, input::LazySet)
     reach = input
     for layer in nnet.layers
         reach = forward_layer(solver, layer, reach)
@@ -14,14 +14,14 @@ end
 # Checks whether the reachable set belongs to the output constraint
 # It is called by all solvers under reachability
 # Note vertices_list is not defined for HPolytope: to be defined
-function check_inclusion(reach::Vector{<:AbstractPolytope}, output)
+function check_inclusion(reach::Vector{<:LazySet}, output)
     for poly in reach
         issubset(poly, output) || return ReachabilityResult(:violated, reach)
     end
     return ReachabilityResult(:holds, similar(reach, 0))
 end
 
-function check_inclusion(reach::P, output) where P<:AbstractPolytope
+function check_inclusion(reach::P, output) where P<:LazySet
     if issubset(reach, output)
         return ReachabilityResult(:holds, P[])
     end
@@ -30,6 +30,8 @@ end
 
 # return a vector so that append! is consistent with the relu forward_partition
 forward_partition(act::Id, input::HPolytope) = [input]
+
+forward_partition(act::Id, input::Zonotope) = input
 
 function forward_partition(act::ReLU, input::HPolytope)
     n = dim(input)
@@ -54,4 +56,9 @@ function getP(h::Int64, n::Int64)
         vec[i] = ifelse(str[i] == '1', 1, 0)
     end
     return Diagonal(vec)
+end
+
+# forward_partition for Zonotopes
+function forward_partition(act::ReLU, input::Zonotope)
+    return overapproximate(Rectification(input), Zonotope)
 end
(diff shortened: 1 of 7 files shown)


src/reachability/utils/reachability.jl
```diff
@@ -3,7 +3,7 @@
 # Performs layer-by-layer propagation
 # It is called by all solvers under reachability
 # TODO: also called by ReluVal and FastLin, so move to general utils (or to network.jl)
-function forward_network(solver, nnet::Network, input::AbstractPolytope)
+function forward_network(solver, nnet::Network, input::LazySet)
```
tomerarnon, line 6: What was the reason to convert all of the `::AbstractPolytope` types into `::LazySet`s instead? Are these functions accepting inputs that aren't AbstractPolytopes? Is it a Rectification or AffineMap?

src/utils/problem.jl
```diff
@@ -108,7 +108,7 @@ Like `BasicResult`, but also returns the output reachable set given the input co
 """
 struct ReachabilityResult <: Result
 	status::Symbol
-	reachable::Vector{<:AbstractPolytope}
+	reachable::typejoin(Array{<:LazySet}, Tuple{<:LazySet})
```
tomerarnon, line 111: `typejoin(Array{<:LazySet}, Tuple{<:LazySet})` is `Any`. I don't mind making this field `Any` in principle, since it has no bearing at all on inference or performance, but I'm also not seeing where a `Tuple{<:LazySet}` is being returned. Was this change temporary during development?

src/reachability/utils/reachability.jl
SebastianGuadalupe: My idea was to let the user pass an input with something like `[1.0, 2.0] × [1.0, 2.0]` but it wasn't done in the end, I will change it back

src/utils/problem.jl
SebastianGuadalupe: I had problems with the solver returning a Tuple but it seems that this is not longer needed!

mforets:
glad to see this moving forward :tada:  thanks to the NV team :+1:

mforets:
for the nonlinear part, the key part is the implementation of `overapproximate(Rectification(input), Zonotope)`, implemented [here](https://github.com/JuliaReach/LazySets.jl/blob/master/src/Approximations/overapproximate.jl#L1640) and which uses a result from Singh. it's probably a good idea to also reference that paper in the new algorithm's docstring.

tomerarnon:
Good point, we will add a reference there. 

Speaking of `Rectification`s, I intend to revisit the reachability algorithms at some opportunity, and implement more lazy implementations alongside the eager versions. I think part of that will be creating different ways of concretizing  `Rectification`s (including maybe precisely as `UnionSetArray`s?). I recall we touched on this idea when we first talked about it many months ago (but can't find that discussion now). Do you have any thoughts on that?

mforets:
> I think part of that will be creating different ways of concretizing Rectifications (including maybe precisely as UnionSetArrays?)

unfortunately the support function of a relu cannot be evaluated efficiently afaik (it was discussed with some length [in this thread](https://github.com/JuliaReach/LazySets.jl/issues/1176) and stuff in that thread could be useful to investigate further.. but i don't have high hopes); hence, one can't lazily work with layers, or do things like iterative refinements of a Rectification efficiently. about concretization using zonotopes as in the new Ai2z implementation, to improve precision one could split the zonotopes along generators, methods which are already available in LazySets and @SebastianGuadaulpe was considering to apply that idea to NV (maybe worth opening an issue here?).

tomerarnon:
> unfortunately the support function of a relu cannot be evaluated efficiently afaik

That's a real shame... Thanks for digging up that thread.

>to improve precision one could split the zonotopes along generators, methods which are already available in LazySets and @SebastianGuadalupe was considering to apply that idea to NV (maybe worth opening an issue here?).

We'd be happy to incorporate further improvements to ai2!

On another matter, does LazySets offer any of way of calculating/approximating the amount of over-approximation of various operations? For example, after over-approximating with some set/algorithm, can I also ever know the % of over-approximation (without computing the exact set too, of course).

mforets:
> On another matter, does LazySets offer any of way of calculating/approximating the amount of over-approximation of various operations?

There is relevant literature from the 90's (see works by Kamenev and also by Lotov), some of which we have implemented  LazySets for the 2D case, see the [Polyhedral approximation of a convex set docs](https://juliareach.github.io/LazySets.jl/dev/man/polyhedral_approximations/#Polyhedral-approximation-of-a-convex-set); it's available in the [LazySets.Approximations.overapproximate](https://juliareach.github.io/LazySets.jl/dev/lib/approximations/#Overapproximations) functions that have an iterative refinement parameter ε. This parameter is an upper bound on the approximation error, measured in terms of the Hausdorff distance. The n-dimensional case, that we haven't implemented, is conceptually the same as in 2D however an efficient implementation would require familiarity with the double description method (see works by Fukuda); relevant open issues are [LazySets#969](https://github.com/JuliaReach/LazySets.jl/issues/969) and [LazySets#1572](https://github.com/JuliaReach/LazySets.jl/issues/1572).

mforets:
> that we haven't implemented, is conceptually the same as in 2D however an efficient implementation would require familiarity

that is to say, a non-efficient implementation (in the sense that it performs extra work by naively computing the whole hrep/vrep at each pass) is doable and i have it in a notebook somewhere...

tomerarnon:
>that is to say, a non-efficient implementation (in the sense that it performs extra work by naively computing the whole hrep/vrep at each pass) is doable and i have it in a notebook somewhere...

This couldn't scale to higher dimensions, so while it is useful for some of our networks, it wouldn't be for most of them.

Good to hear though that there are open issues on the matter already!

PR #133, opened 2020-07-28
Combine ai2

tomerarnon:
@SebastianGuadalupe here I try to consolidate the 3 methods of Ai2 into one Ai2{T}


src/reachability/ai2.jl
```diff
@@ -18,25 +32,49 @@ Reachability analysis using split and join.
...
 
+# method for Zonotope and Hyperrectangle, if the input set isn't a Zonotope
+function forward_layer(solver::Ai2, L::Layer, input::AbstractPolytope)
+    X = overapproximate(input, Hyperrectangle)
+    return forward_layer(solver, L, X)
+end
+
+function forward_layer(solver::Ai2{T}, L::Layer, input::AbstractZonotope) where T<:Union{Hyperrectangle, Zonotope}
+    Ẑ = affine_map(L, input)
+    return overapproximate(Rectification(Ẑ), T)
```
SebastianGuadalupe, line 74: Note that `rectify(Ẑ)` is faster in the case of the Hyperrectangle, as `overapproximate(Rectification(Ẑ), Hyperrectangle)` does some extra computations that are not needed

src/reachability/ai2.jl
```diff
@@ -18,25 +32,49 @@ Reachability analysis using split and join.
...
+end
+
+
 # extend lazysets convex_hull to a vector of polytopes
 function LazySets.convex_hull(sets::Vector{<:AbstractPolytope}; backend = CDDLib.Library())
```
SebastianGuadalupe, line 79: In LazySets there is a function that does this: https://github.com/JuliaReach/LazySets.jl/pull/2215

src/reachability/ai2.jl
SebastianGuadalupe: It's not that much faster, so it may not be worth the extra code, it's up to you
```julia
julia> H = rand(Hyperrectangle, dim=100);

julia> @btime rectify($H);
  4.720 ns (9 allocations: 5.31 KiB)

julia> @btime overapproximate(Rectification($H), Hyperrectangle);
  29.632 ns (10 allocations: 5.34 KiB)
```

src/reachability/ai2.jl
tomerarnon: Not quite. It returns a vector of vertices and we want a Polytope (we could do VPolytope(...) at the end though). Also, in hrep, I think it's more efficient not to go to vrep, no?

SebastianGuadalupe: The problem here is that you are doing a `convex_hull` every loop, while that implementation only does once, so, if there are more than two polytopes, I think that it will be faster even after the conversion, and it will scale much better with more polytopes and dimensions.

tomerarnon: Hmm, this could be. I didn't think so since I assumed `convex_hull` of two polytopes of the same type does not convert, and therefore it should be more efficient. Since the code in LazySets defers to Polyhedra, do you know if this is the case or not?

src/reachability/ai2.jl
tomerarnon: I don't mind another 1 line method definition 😄 
I'll add it in a moment.

tomerarnon:
Hmm... It seems tests are failing due to Box or Ai2z being too overapproximate. But in your PRs the tests passed just fine. Do you see anywhere where I made a mistake @SebastianGuadalupe?

src/reachability/ai2.jl
SebastianGuadalupe: The function used is `removehredundancy()` [here](https://github.com/JuliaReach/LazySets.jl/blob/77e9f6c30f8f49fbde6fb8137a0ddfcd820eda9b/src/Sets/HPolyhedron.jl#L451), so it does not convert to v-rep.
I did a little benchmark with two polygons:
```julia
julia> @btime convert(HPolytope, VPolytope(convex_hull(UnionSetArray([$H, $H]))));
  331.000 ns (4620 allocations: 310.05 KiB)
julia> @btime convex_hull($H, $H);
  1.072 ms (14202 allocations: 864.63 KiB)
```
So it seems to me that the implementation in LazySets is faster even in 2D and two polytopes

tomerarnon: Computing `Pch` in Polyhedra requires vrep I think. The Polyhedra documentation suggests it is only possible in vrep
https://juliapolyhedra.github.io/Polyhedra.jl/latest/utilities/#Polyhedra.convexhull

Therefore your method is better after all I guess

tomerarnon: By the way, I have recently noticed printouts from the LP solver used by LazySets (or lower in the stack) which were not there in prior package versions. Do you know which issue/s these new messages are related to? An example is:
```julia
julia> convex_hull(UnionSetArray([hp, hp]))
glp_simplex: unable to recover undefined or non-optimal solution
glp_simplex: unable to recover undefined or non-optimal solution
glp_simplex: unable to recover undefined or non-optimal solution
glp_simplex: unable to recover undefined or non-optimal solution
glp_simplex: unable to recover undefined or non-optimal solution
glp_simplex: unable to recover undefined or non-optimal solution
3-element Array{Array{Float64,1},1}:
 [-0.36626701868929534, 1.1382211150524333, -0.6787710912863433]
 [0.36968831822766857, -0.7100424633428055, -0.1920207505884401]
 [-1.9245657296871925, 0.42914282672330284, -0.7531344058289536]
```

SebastianGuadalupe: I don't know where they originate but there are issues that mentions it(https://github.com/JuliaReach/LazySets.jl/issues/2278#issuecomment-664180892 and https://github.com/JuliaReach/LazySets.jl/issues/861#issuecomment-660503226)

SebastianGuadalupe:
> Hmm... It seems tests are failing due to Box or Ai2z being too overapproximate (returning :violated for :holds). But in your PRs the tests passed just fine. Do you see anywhere where I made a mistake @SebastianGuadalupe?

I run the test locally and it doesn't seems to be a problem of being too overapproximate, the resulting set does not include the expected result, my guess is that the problem is in the affine map?

tomerarnon:
> my guess is that the problem is in the affine map?

I believe you also used affine_map in Ai2z, no? In the travis tests, that is the first failure case encountered. Very strange indeed!

tomerarnon:
I think I found the issue... I unwittingly removed definitions for identity activation. I'll check on my end if that's it.

SebastianGuadalupe:
Yes, that is the error, I just found it too

SebastianGuadalupe:
Great to see `ReachabilityResult` always returning a reach set, I wondered why it was not the case, we have applications where it is very useful!

tomerarnon:
Indeed, I consider that to be a design mistake. I figured I might as well fix it here than in another PR 🤷‍♂️

PR #146, opened 2020-08-03
maxsens: resolve integer overflow

### LINKED ISSUE #140: MaxSens resolution not working
When debugging for the error, it seems to me that is in the function `partition(input::Hyperrectangle, Δ)`, printing the set before and after, the set is empty after the function is used. When using `solve(MaxSens(resolution = 0), problem_mnist)`, the algorithm works fine, so `Δ == 0 && return [input]` it's doing its job and can be used as a workaround.

tomerarnon:
fixes #140


tomerarnon:
It takes the product of `n_hypers_per_dim`, first converting each number into a `BigInt` with `big`. This means the product isn't sensitive to integer overflow, which was the problem in #140. Before, both `prod` and `length(CI)` were subject to overflow.
E.g.
```julia
julia> n = fill(2, 65);

julia> prod(n)
0

julia> prod(big, n)
36893488147419103232
```

PR #152, opened 2020-08-05
Add relax_ReLU to util

changliuliu:
This branch is built on top of #79 . The function relax_ReLU is used in both ConvDual and Neurify (was the calc_slop function).


src/utils/util.jl
```diff
@@ -229,6 +229,25 @@ Currently only support ReLU and Id.
...
+
+Returns the slope of a ReLU activation based on its lower and upper bounds
+
+Inputs:
+- `l::Float64`: lower bound
+- `u::Float64`: upper bound
+Return:
+- `slop::Float64`: 0 if u<0; 1 if l>0; u/(u-l) otherwise
+"""
+function act_gradient(act::ReLU, l::Float64, u::Float64)
```
tomerarnon, line 243: This signature + behavior might be misleading. Technically this isn't the activation gradient of ReLU, but of the triangle relaxation. In that sense it is different from the other `act_gradient` methods, which are not approximate. Should we rename it to `relaxed_relu_gradient(l, u)`, for example, or something else more accurate?

changliuliu: yes, that is a good idea.

tomerarnon:
I moved only the two commits related to `act_gradient` over to the neurify branch directly using `git cherry-pick`. Closing this PR as a result.

Issue #149, opened 2020-08-03
NSVerify and ILP disagreement - likely ILP unsoundness.


castrong:
**With Code:**
```
# Setup problem
network = NeuralVerification.read_nnet("./examples/networks/issue_net.nnet")
input_set = LazySets.Hyperrectangle([0.3239560971706661, -0.051023198756482646, 0.10643884307283402, 0.2617471526666566, -0.4876992062506713], [0.27511191847549554, 0.4726664814672026, 0.4758823479375396, 0.3623388331604731, 0.22049620536613213])
output_set = NeuralVerification.PolytopeComplement(LazySets.HPolytope([LazySets.HalfSpace([1.0], 0.46396369980702556), LazySets.HalfSpace([-1.0], -0.4508776562797363)]))
problem = NeuralVerification.Problem(network, input_set, output_set)

ns_status = NeuralVerification.solve(NeuralVerification.NSVerify(), problem).status
ilp_status =  NeuralVerification.solve(NeuralVerification.ILP(), problem).status
println("NSVerify status: ", ns_status)
println("ILP status: ", ilp_status)

```

**And Network: **
//Neural Network File Format by Kyle Julian, Stanford 2016.
2, 1, 1, 8,
1, 8, 1,
This line extraneous
-6.55e4,
6.55e4,
0.0,0.0,
1.0,1.0,
0.8678491028912014, -0.5203780903607784, -0.5142031406625782, -0.6977040448276313, -0.976885080551285,
-0.1460743574085881, 0.3154578548182978, 0.9532183372024212, -0.574760350454679, 0.6341150831930098,
-0.7898630677526377, -0.7313180250715745, -0.20467484308960016, 0.40858515372716386, -0.6385823044294656,
-0.35203560829994407, 0.3166304799264661, -0.4619174469634739, 0.9392249552449465, 0.9352413917562559,
-0.5166440988054104, -0.7696079370706981, -0.6942957067572171, 0.7201009053603964, -0.007367629464616776,
-0.11301221277218598, -0.8316575309519427, -0.5740001481764962, 0.585516746004052, -0.1275076273402127,
0.6528733981400743, -0.528579455088038, 0.5254640882299264, -0.18236568211966997, 0.4216232004175584,
0.09289100811092821, -0.5906110198701051, -0.294992020863452, -0.7075907072711356, 0.05897916487234678,
-0.7916178249103649,
0.8827517443779862,
0.9343433076177265,
0.6578225443695569,
[...]

PR #154, opened 2020-08-06
Fix iLP #149

changliuliu:
iLP only checks one linear branch of the network that has the same activation as the center of the input set. Hence it should return false for any input set that goes beyond one linear branch.

Now instead of comparing the maximum allowable disturbance with the maximum input radius (scalar to scalar comparison), we check the possible deviation in every dimension (vector to vector comparison). If in either dimension, the input radius is greater than the allowable disturbance in that dimension, we will return a :violation.

This pull request also simplifies MIPVerify. Since we directly encode the input constraints in the problem, it is not possible for the maximum allowable disturbance to be greater than the maximum input radius. Hence if there is a solution to the optimization problem, then the solution must be a counter example. In this way, MIPVerify is sound and complete.

### DIFF
--- a/src/optimization/iLP.jl
+++ b/src/optimization/iLP.jl
@@ -63,10 +63,12 @@ function solve(solver::ILP, problem::Problem)
 end
 
 function interpret_result(solver::ILP, x, input)
-    if prod(abs.(x - input.center) .>= input.radius)
-        return AdversarialResult(:holds, minimum(abs.(x - input.center)))
+    radius = abs.(x .- center(input))
+
+    if all(radius .>= radius_hyperrectangle(input))
+        return AdversarialResult(:holds, minimum(radius))
     else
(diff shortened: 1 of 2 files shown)


src/optimization/iLP.jl
```diff
@@ -55,17 +56,17 @@ function solve(solver::ILP, problem::Problem)
...
         end
         add_constraint!(model, nnet, neurons, δ, index)
     end
 end
 
-function interpret_result(solver::ILP, o, input)
-    if o >= maximum(input.radius)
-        return AdversarialResult(:holds, o)
+function interpret_result(solver::ILP, x, input)
+    if prod(abs.(x - input.center) .>= input.radius)
```
tomerarnon, line 66: This is false if any of the lhs < rhs. So we can do 
`all(abs.(x - input.center) .>= input.radius)`

PR #163, opened 2020-08-26
Bounds plotting

castrong:
Changes to several methods to get access to intermediate bounds. Plotting code that visualizes the differences between bounds.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -13,14 +13,15 @@ Parameters = "d96e819e-fc66-5662-9728-84c9c7592b0a"
 PicoSAT = "ff2beb65-d7cd-5ff1-a187-74671133a339"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 Polyhedra = "67491407-f73d-577b-9b50-8179a7c68029"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SCS = "c946c3f1-0d1f-5ce8-9dea-7daa1f7e2d13"
 
 [extras]
 Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["Test", "Flux", "JLD2"]
--- a/src/adversarial/neurify.jl
+++ b/src/adversarial/neurify.jl
@@ -53,7 +53,9 @@ function solve(solver::Neurify, problem::Problem)
             domain, splits = select!(reach_list, solver.tree_search)
         end
 
-        reach = forward_network(solver, nnet, domain)
+        reach = forward_network(solver, nnet, domain, collect=true)
+        # The first entry is the input set
+        popfirst!(reach)
         result, max_violation_con = check_inclusion(solver, nnet, last(reach).sym, output)
 
         if result.status === :violated
@@ -195,33 +197,18 @@ function get_max_nodewise_influence(nnet::Network,
     return (i_max, j_max, influence_max)
 end
 
-
-
-function forward_network(solver::Neurify, network::Network, input)
-    forward_network(solver, network, init_symbolic_grad(input))
-end
-function forward_network(solver::Neurify, network::Network, input::SymbolicIntervalGradient)
-    reachable = [input = forward_layer(solver, L, input) for L in network.layers]
-    return reachable
-end
-
-
-function forward_layer(solver::Neurify, layer::Layer, input)
-    return forward_act(solver, forward_linear(solver, input, layer), layer)
-end
-
 # Symbolic forward_linear
-function forward_linear(solver::Neurify, input::SymbolicIntervalGradient, layer::Layer)
-    output_Low, output_Up = interval_map(layer.weights, input.sym.Low, input.sym.Up)
-    output_Up[:, end] += layer.bias
-    output_Low[:, end] += layer.bias
+function forward_linear(solver::Neurify, L::Layer, input::SymbolicIntervalGradient)
+    output_Low, output_Up = interval_map(L.weights, input.sym.Low, input.sym.Up)
+    output_Up[:, end] += L.bias
+    output_Low[:, end] += L.bias
     sym = SymbolicInterval(output_Low, output_Up, domain(input))
     return SymbolicIntervalGradient(sym, input.LΛ, input.UΛ)
 end
 
 # Symbolic forward_act
-function forward_act(solver::Neurify, input::SymbolicIntervalGradient, layer::Layer{ReLU})
-    n_node = n_nodes(layer)
+function forward_act(solver::Neurify, L::Layer{ReLU}, input::SymbolicIntervalGradient)
+    n_node = n_nodes(L)
     output_Low, output_Up = copy(input.sym.Low), copy(input.sym.Up)
     LΛᵢ, UΛᵢ = zeros(n_node), ones(n_node)
     # Symbolic linear relaxation
@@ -246,8 +233,8 @@ function forward_act(solver::Neurify, input::SymbolicIntervalGradient, layer::La
     return SymbolicIntervalGradient(sym, LΛ, UΛ)
 end
 
-function forward_act(solver::Neurify, input::SymbolicIntervalGradient, layer::Layer{Id})
-    n_node = n_nodes(layer)
+function forward_act(solver::Neurify, L::Layer{Id}, input::SymbolicIntervalGradient)
+    n_node = n_nodes(L)
     LΛ = push!(input.LΛ, ones(n_node))
     UΛ = push!(input.UΛ, ones(n_node))
     return SymbolicIntervalGradient(input.sym, LΛ, UΛ)
--- a/src/optimization/convDual.jl
+++ b/src/optimization/convDual.jl
@@ -72,6 +72,19 @@ function backprop!(v::Vector{Float64}, u::Vector{Float64}, l::Vector{Float64})
     return o
 end
 
+function get_bounds(solver::ConvDual, nnet::Network, input::Hyperrectangle; before_act::Bool=true)
+    # only supports pre-activation bounds for now
+    @assert before_act == true "ConvDual get_bounds only currently supports pre-activation bounds"
+    # only supports hypercubes in this implementation
+    @assert all(input.radius[1] .== input.radius)
+    L, U = get_bounds(nnet, input.center, input.radius[1])
+    # add the bounds from the input set
+    pushfirst!(L, low(input))
+    pushfirst!(U, high(input))
+    # convert convdual's bounds into hyperrectangles
+    return [Hyperrectangle(low=L[i], high=U[i]) for i = 1:length(L)]
+end
+
 # Forward_network and forward_layer:
 # This step is similar to reachability method
 function get_bounds(nnet::Network, input::Vector{Float64}, ϵ::Float64)
--- a/src/reachability/ai2.jl
+++ b/src/reachability/ai2.jl
@@ -55,34 +55,22 @@ function solve(solver::Ai2, problem::Problem)
     return check_inclusion(reach, problem.output)
 end
 
-forward_layer(solver::Ai2, L::Layer, inputs::Vector) = forward_layer.(solver, L, inputs)
+# Ai2h and Ai2z use affine_map
+# Box uses approximate_affine_map for the linear region if it is propagating a zonotope
+forward_linear(solver::Ai2h, L::Layer{ReLU}, input::AbstractPolytope) = affine_map(L, input)
+forward_linear(solver::Ai2z, L::Layer{ReLU}, input::AbstractZonotope) = affine_map(L, input)
+forward_linear(solver::Box, L::Layer{ReLU}, input::AbstractZonotope) = approximate_affine_map(L, input)
+# method for Zonotope and Hyperrectangle, if the input set isn't a Zonotope overapproximate
+forward_linear(solver::Union{Ai2z, Box}, L::Layer{ReLU}, input::AbstractPolytope) = forward_linear(solver, L, overapproximate(input, Hyperrectangle))
 
-function forward_layer(solver::Ai2h, L::Layer{ReLU}, input::AbstractPolytope)
-    Ẑ = affine_map(L, input)
-    relued_subsets = forward_partition(L.activation, Ẑ) # defined in reachability.jl
-    return convex_hull(UnionSetArray(relued_subsets))
-end
+# Forward_act is different for Ai2h, Ai2z and Box
+forward_act(solver::Ai2h, L::Layer{ReLU}, Ẑ::AbstractPolytope) = convex_hull(UnionSetArray(forward_partition(L.activation, Ẑ)))
+forward_act(solver::Ai2z, L::Layer{ReLU}, Ẑ::AbstractZonotope) = overapproximate(Rectification(Ẑ), Zonotope)
+forward_act(slver::Box, L::Layer{ReLU}, Ẑ::AbstractPolytope) = rectify(Ẑ)
 
-# method for Zonotope and Hyperrectangle, if the input set isn't a Zonotope
-function forward_layer(solver::Union{Ai2z, Box}, L::Layer{ReLU}, input::AbstractPolytope)
-    return forward_layer(solver, L, overapproximate(input, Hyperrectangle))
-end
-
-function forward_layer(solver::Ai2z, L::Layer{ReLU}, input::AbstractZonotope)
-    Ẑ = affine_map(L, input)
-    return overapproximate(Rectification(Ẑ), Zonotope)
-end
-
-
-function forward_layer(solver::Box, L::Layer{ReLU}, input::AbstractZonotope)
-    Ẑ = approximate_affine_map(L, input)
-    return rectify(Ẑ)
-end
-
-function forward_layer(solver::Ai2, L::Layer{Id}, input)
-    return affine_map(L, input)
-end
-
+# For ID activation do an affine map for all methods
+forward_linear(solver::Ai2, L::Layer{Id}, input::AbstractPolytope) = affine_map(L, input)
+forward_act(solver::Ai2, L::Layer{Id}, input::AbstractPolytope) = input
 
 function convex_hull(U::UnionSetArray{<:Any, <:HPolytope})
     tohrep(VPolytope(LazySets.convex_hull(U)))
--- a/src/reachability/exactReach.jl
+++ b/src/reachability/exactReach.jl
@@ -29,18 +29,19 @@ function solve(solver::ExactReach, problem::Problem)
     return check_inclusion(reach, problem.output)
 end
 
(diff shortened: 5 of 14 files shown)


src/reachability/utils/reachability.jl
```diff
@@ -3,12 +3,51 @@
...
+                                  T=Any)
+    zs = (input, input)
+    collection::Vector{T} = [collect_f(zs)]
+    for L in nnet.layers
+        zs = forward_layer(solver, L, zs[2])
+        push!(collection, collect_f(zs))
+    end
+    return collection
+end
+function get_bounds(solver, nnet, input; before_act=false)
```
tomerarnon, line 40: Since we're adding a `get_bounds(solver, nnet, input; before_act)`, we should probably change the definition of the original (IA) get_bounds to have the same keyword, rather than optional argument act. Currently it is `get_bounds(nnet, input[, act])`

src/reachability/utils/reachability.jl
```diff
@@ -3,12 +3,51 @@
...
+    f = before_act ? first : last
+    _forward_network_collect(solver, nnet, input,
+                             set->overapproximate(f(set), Hyperrectangle),
+                             Hyperrectangle)
```
tomerarnon, line 44: There is a typo here. Probably you meant for the last Hyperrectangle to not be there?

src/reachability/utils/reachability.jl
```diff
@@ -3,12 +3,51 @@
 # Performs layer-by-layer propagation
 # It is called by all solvers under reachability
 # TODO: also called by ReluVal and FastLin, so move to general utils (or to network.jl)
-function forward_network(solver, nnet::Network, input)
-    reach = input
+function forward_network(solver::Solver, nnet::Network, input;
+                         collect=false, before_act=false, overapprox=false, T=Any)
```
tomerarnon, line 7: Instead of defaulting to T = Any maybe we should implement a way of inferring the return type of forward_layer? At a low level, a lot of LazySets function are type unstable in order to return e.g. EmptySet vs HPolytope, but at the level of forward_layer we are generally careful with out returns, so we might have decent type stability we can use with function like Base.return_types or Base._return_type. For one thing, this avoids the ugly T = SymbolicInteralGradient in neurify.

src/NeuralVerification.jl
```diff
@@ -100,4 +101,8 @@ const TOL = Ref(sqrt(eps()))
 set_tolerance(x::Real) = (TOL[] = x)
 export set_tolerance
 
+LazySets.overapproximate(set::Vector{<:HPolytope}, ::Type{Hyperrectangle}) = overapproximate(overapproximate.(set, Hyperrectangle), Hyperrectangle)
+LazySets.overapproximate(set::Vector{<:AbstractZonotope}, ::Type{Hyperrectangle}) = overapproximate(ConvexHullArray(set), Hyperrectangle)
+LazySets.overapproximate(set::Union{SymbolicIntervalMask, SymbolicIntervalGradient}, ::Type{Hyperrectangle}) = Hyperrectangle(low=low(set), high=high(set))
+
```
tomerarnon, line 107: These are type piracy (except the last one, where we own the type). Let's figure out a way to accomplish what these are trying to do without defining them this way. What is the major reason for the first two?

src/adversarial/neurify.jl
```diff
@@ -195,33 +195,23 @@ function get_max_nodewise_influence(nnet::Network,
...
-    reachable = [input = forward_layer(solver, L, input) for L in network.layers]
-    return reachable
-end
-
-
-function forward_layer(solver::Neurify, layer::Layer, input)
-    return forward_act(solver, forward_linear(solver, input, layer), layer)
+function forward_network(solver::Neurify, network::Network, input::HPolytope)
+    println("in Neurify forward network")
+    return forward_network(solver, network, init_symbolic_grad(input))
```
tomerarnon, line 200: If we're going to redo forward_network as we are, let's get rid of definitions like these. We can do init_symbolic_grad directly in solve like we do for the first domain, and eliminate the need for this.

There is also a leftover debug statement 😉

src/optimization/convDual.jl
```diff
@@ -72,6 +72,17 @@ function backprop!(v::Vector{Float64}, u::Vector{Float64}, l::Vector{Float64})
...
+function get_bounds(solver::ConvDual, nnet::Network, input::Hyperrectangle)
+    # only supports hypercubes in this implementation
+    @assert all(input.radius[1] .== input.radius)
+    L, U = get_bounds(nnet, input.center, input.radius[1])
+    # add the bounds from the input set
+    pushfirst!(L, low(input))
+    pushfirst!(U, high(input))
+    # convert convdual's bounds into hyperrectangles
+    return [Hyperrectangle(low=L[i], high=U[i]) for i = 1:length(L)]
+end
```
tomerarnon, line 84: This doesn't yet stick to the new `get_bounds` interface, since it doesn't support `before_act=true/false`. Is there a way that it could? If not, perhaps just adding the keyword as a dummy, and then asserting that it is `false` would make it at least behave the same.

src/reachability/ai2.jl
```diff
@@ -55,34 +55,22 @@ function solve(solver::Ai2, problem::Problem)
...
-    return rectify(Ẑ)
-end
-
-function forward_layer(solver::Ai2, L::Layer{Id}, input)
-    return affine_map(L, input)
-end
-
+# Ai2h and Ai2z use affine_map
+# Box uses approximate_affine_map for the linear region if it is propagating a zonotope
+forward_linear(solver::Union{Ai2h}, L::Layer{ReLU}, input::AbstractPolytope) = affine_map(L, input)
```
tomerarnon, line 60: Single union types generally don't make sense... Is there something you were trying to accomplish here?

src/reachability/exactReach.jl
```diff
@@ -29,18 +29,19 @@ function solve(solver::ExactReach, problem::Problem)
...
+forward_linear(solver::ExactReach, L::Layer, input) = forward_linear(solver, L, convert(HPolytope, input))
 
-function forward_layer(solver::ExactReach, layer::Layer, input::Vector{<:HPolytope})
+# Forward linear for a single polytope and a list of polytopes
+forward_linear(solver::ExactReach, L::Layer, input::HPolytope) = affine_map(L, input)
+forward_linear(solver::ExactReach, L::Layer, input::Vector{<:HPolytope}) = [affine_map(L, set) for set in input]
+
+# Forward act for a single polytope and a list of polytopes
+forward_act(solver::ExactReach, L::Layer, input::HPolytope) = forward_partition(L.activation, input)
+function forward_act(solver::ExactReach, L::Layer, input::Vector{<:HPolytope})
```
tomerarnon, line 41: This is the major downside of splitting `forward_layer` to always be `forward_linear` followed by `forward_act`. We allocate a lot of unnecessary stuff here. Maybe there is something broadcast-y we can do in the definition or call of forward_layer that would help mitigate this...

src/satisfiability/planet.jl
```diff
@@ -53,14 +53,13 @@ function init_ψ(nnet::Network, bounds::Vector{Hyperrectangle})
     ψ = Vector{Vector{Int64}}()
     index = 0
     for i in 1:length(bounds)-1
-        index = set_activation_pattern!(ψ, nnet.layers[i], bounds[i], index)
+        index = set_activation_pattern!(ψ, nnet.layers[i], bounds[i+1], index)
```
tomerarnon, line 56: Is this i+1 the same thing as in neurify?

src/satisfiability/planet.jl
```diff
@@ -132,31 +131,39 @@ function max_slack(x::Vector{<:Vector}, act)
...
-# Only use tighten_bounds for feasibility check
-function tighten_bounds(problem::Problem, optimizer)
+function tighten_bounds(problem::Problem, optimizer;
+                        bounds::Vector{Hyperrectangle} = get_bounds(problem, false),
+                        encoding = TriangularRelaxedLP())
+    new_bounds = copy(bounds)
     model = Model(optimizer)
-    model[:bounds] = bounds = get_bounds(problem)
-    z = init_vars(model, problem.network, :z, with_input=true)
+    init_model_vars(model, problem, encoding, bounds=new_bounds, before_act=true)
```
tomerarnon, line 139: alternatively, we could have `bounds = copy(bounds)`, and then `new_bounds =model[:bounds]` after. Whichever is clearer.

src/utils/util.jl
```diff
@@ -183,15 +183,14 @@ return Vector{Vector{Int64}}.
...
     end
     return act_pattern
 end
 
 function get_activation(L::Layer{ReLU}, bounds::Hyperrectangle)
-    before_act_bound = approximate_affine_map(L, bounds)
-    lower = low(before_act_bound)
-    upper = high(before_act_bound)
+    lower = low(bounds)
+    upper = high(bounds)
```
tomerarnon, line 193: This is a dangerous change... this function is used in lots of places. You're sure it doesn't have ill effects?

Also lets update the docstring and maybe the variable name to reflect these are pre-activation bounds.

tomerarnon (changes requested):
I've left some comments on some things that need to be cleaned up a bit before merging.

An additional comment at a high level: the recent changes in favor of pre activation bounds has made it ambiguous in some places what a variable like `bounds` refers to, and we should be more specific. Sometimes it doesn't matter (e.g. `tighten_bounds`, where everything is magic) but sometimes it does (e.g. `get_activation`, where the wrong interpretation will break a lot downstream). In addition, although we have taken it for granted that pre-activation is "better" than post, we have made no moves to test or prove this, and post-activation should actually be tighter, so it may be the other way around, or there may be no difference at all. At some point that test should be conducted to settle that (I don't volunteer to do it for another 10 months 😅).

tomerarnon:
The testing failure was coming from `Random` not being in the dependencies (it's not in Project.toml). Fixing that leads to tighten_bounds failing at `encode_layer` due to indexing convention (should be i-1)... I expect this will come up again an again until we sort out bounds for good (see my review comment).

src/reachability/utils/reachability.jl
tomerarnon: My mistake; didn't realize the first Hyperrectangle was inside the overapproximate

src/reachability/utils/reachability.jl
castrong: `get_bounds(problem::Problem, args...) = get_bounds(problem.network, problem.input, args...)
`
Is there a good way to handle this line if I make the original get_bounds require a keyword? I'm not too familiar with the args... and how I can make it work

tomerarnon: You can forward both regular and keyword arguments if that's what you mean. E.g. you can have something like:
```julia
f(x, y; z = 1) = x + y + z
f(x, args...; kwargs...) = f(x, sum(args); kwargs...)

f(1, 2, 3, 4, 5, z = 6)  # 21
```

src/NeuralVerification.jl
castrong: Hmm okay - it's so that we can always use overapproximate(set, Hyperrectangle) in our get_bounds function in reachability.jl. I'm not sure the best way to keep that function generate without overloading overapproximate so that it can work for any oof our approximate reachable sets.

castrong: Would it be better to create our own overapproximate function (with a different name) which we define for some types of inputs by calling to LazySets and for others with these implementations? we could call it "overapproximate_all" or something?

src/adversarial/neurify.jl
castrong: oh so we wouldn't even need this function? Just anywhere we call forward network we have to call init_symbolic_grad first?

src/satisfiability/planet.jl
castrong: I don't think I follow - which part of neurify?. I changed it to i+1  because we want it to correpsond to the layer we're indexing into. The first layer corresponds to the second index of bounds since we don't have a layer object corresponding to the input layer.

src/satisfiability/planet.jl
castrong: oh because we'll be updating them inside the model and we can just pull those out at the end?

castrong: I tried updating this, I think it makes it cleaner

src/utils/util.jl
castrong: I believe it should be equivalent to what we had before? Before we had the approximate affine map happening in bounds. Now I was trying to consistently use pre-activation bounds, and so I used them here as well but I think it effectively is just switching where approximate affine map is getting applied? Unless I'm misinterpreting this? Could it be a different approximate affine map that gets used? Although it was already being used on bounds, just on post-activation bounds, so we're still dealing with approximate affine maps on hyperrectangles.

castrong: I think if it is equivalent it might be cleaner this way, since we don't have to do the approximate affine map again ( in the old version it's already being done in get_bounds when we get post-activation bounds, then we apply it again in this function).

castrong: The only place I found where get_bounds is called with bounds passed in is in Planet. Since I updated Planet to use pre-activation bounds,  this should be okay. All other uses of get_activation seem to pass in a single point, so I think those should still be ok.

castrong: I updated the doc string - in general I don't know what to do between pre and post activation bounds.

src/adversarial/neurify.jl
```diff
@@ -53,7 +53,7 @@ function solve(solver::Neurify, problem::Problem)
             domain, splits = select!(reach_list, solver.tree_search)
         end
 
-        reach = forward_network(solver, nnet, domain)
+        reach = forward_network(solver, nnet, domain; collect=true, T=SymbolicIntervalGradient)[1:end]
```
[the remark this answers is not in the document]
castrong: do we even need to popfirst? I think it works now (I had updated indexing at some point) with just keeping the full set - I'll push a version without the popfirst

src/adversarial/neurify.jl
```diff
@@ -125,8 +125,8 @@ function constraint_refinement(nnet::Network,
     i, j, influence = get_max_nodewise_influence(nnet, reach, max_violation_con, splits)
     # We can generate three more constraints
     # Symbolic representation of node i j is Low[i][j,:] and Up[i][j,:]
-    aL, bL = reach[i].sym.Low[j, 1:end-1], reach[i].sym.Low[j, end]
-    aU, bU = reach[i].sym.Up[j, 1:end-1], reach[i].sym.Up[j, end]
+    aL, bL = reach[i+1].sym.Low[j, 1:end-1], reach[i+1].sym.Low[j, end]
```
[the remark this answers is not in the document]
castrong: Ok, I updated so that it's i instead of i+1 - is that what was confusing?

src/NeuralVerification.jl
castrong: Okay I just made our own reach_overapprox method in reachability.jl - let me know what you think, it at least resolves the piracy thing.

src/satisfiability/planet.jl
castrong: Oh like is it from the input having bounds? I believe so

src/reachability/exactReach.jl
castrong: Yeah I'm struggling to come up with a solution, I'm not too familiar with broadcasting.

tomerarnon (changes requested):

src/reachability/utils/reachability.jl
```diff
@@ -27,24 +27,24 @@ end
...
+    T = typeof(elem)
+    collection = T[elem]
     for L in nnet.layers
         zs = forward_layer(solver, L, zs[2])
         push!(collection, collect_f(zs))
     end
     return collection
 end
-function get_bounds(solver, nnet, input; before_act=false)
+function get_bounds(solver::Solver, nnet::Network, input; before_act::Bool=false)
```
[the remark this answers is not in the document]
castrong: Oh, I thought it was good practice in general to annotate them when you could? I had no reason other than that.

src/reachability/utils/reachability.jl
```diff
@@ -153,3 +151,9 @@ LazySets.low(sym::_SymIntOrGrad) = [lower_bound(lower(sym), j) for j in 1:dim(sy
 function radius(sym::_SymIntOrGrad, j::Integer)
     upper_bound(upper(sym), j) - lower_bound(lower(sym), j)
 end
+
+####### Hyperrectangle overapproximations for getting bounds ########
+overapprox_reach(set::Vector{<:HPolytope}, ::Type{Hyperrectangle}) = overapproximate(overapproximate.(set, Hyperrectangle), Hyperrectangle)
+overapprox_reach(set::Vector{<:AbstractZonotope}, ::Type{Hyperrectangle}) = overapproximate(ConvexHullArray(set), Hyperrectangle)
+overapprox_reach(set::Union{SymbolicIntervalMask, SymbolicIntervalGradient}, ::Type{Hyperrectangle}) = Hyperrectangle(low=low(set), high=high(set))
+overapprox_reach(set, ::Type{Hyperrectangle}) = LazySets.overapproximate(set, Hyperrectangle)
```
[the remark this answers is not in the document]
castrong: Which one? Sorry it shows up with all 4 highlighted.

Project.toml
```diff
@@ -6,21 +6,23 @@ CDDLib = "3391f64e-dcde-5f30-b752-e11513730f60"
 GLPK = "60bf3e95-4087-53dc-ae20-288a0d20c6a6"
 GLPKMathProgInterface = "3c7084bd-78ad-589a-b5bb-dbd673274bea"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
+JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
```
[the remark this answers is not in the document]
castrong: The new test mnist_1000.jl uses JLD2 so in order to run the tests I needed to add it.

src/reachability/exactReach.jl
castrong: Ok for now we'll live with this double allocation, so I'll resolve it for now

src/reachability/utils/reachability.jl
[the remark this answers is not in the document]
castrong: I thought it was always better to annotate if you could, I had no specific reason for it

src/reachability/utils/reachability.jl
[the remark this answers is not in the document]
castrong: I'm not sure I understand what you're picturing - I just pushed a new version where I tried to make use of ConvexHullArray for any case where we have a vector of sets getting passed in.

src/reachability/utils/reachability.jl
```diff
@@ -27,24 +27,24 @@ end
...
                                   nnet::Network,
                                   input,
-                                  collect_f=last,
-                                  T=Any)
+                                  collect_f=last)
     zs = (input, input)
-    collection::Vector{T} = [collect_f(zs)]
+    elem = collect_f(zs)
+    T = typeof(elem)
+    collection = T[elem]
```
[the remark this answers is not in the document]
castrong: For bounds we end up with the same type b/c collect_f will always return a Hyperrectangle. If we don't assume that collect_f always leads to the same type then I'm still not sure what to do here. I tried some to get collect_f to tell us the type it will give back, but we don't know at the start what type we'll be inputting into it

src/reachability/utils/reachability.jl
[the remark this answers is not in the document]
tomerarnon: Quite the opposite. It is better not to annotate when possible, and in general always best to annotate with the broadest possible type. In this case, it's true that `solver` and `network` will always have those types, but that's exactly the reason that annotating serves no purpose here

PR #167, opened 2020-08-31
Fix opt2

### LINKED ISSUE #71: JuMP return code mishandling leading to inconsistency

tomerarnon:
I think this closes #71 for good, in addition to previous changes that bounded inputs for some algorithms. Supercedes #72.

### DIFF
--- a/src/optimization/mipVerify.jl
+++ b/src/optimization/mipVerify.jl
@@ -43,8 +43,8 @@ function solve(solver::MIPVerify, problem::Problem)
     encode_network!(model, problem.network, BoundedMixedIntegerLP())
     o = max_disturbance!(model, first(z) - problem.input.center)
     optimize!(model)
-    if termination_status(model) == INFEASIBLE
-        return AdversarialResult(:holds)
+    if termination_status(model) == OPTIMAL
+        return AdversarialResult(:violated, value(o))
     end
-    return AdversarialResult(:violated, value(o))
+    return AdversarialResult(:holds)
 end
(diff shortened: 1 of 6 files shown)


tomerarnon:
superceded by #168

PR #168, opened 2020-08-31
Type piracy

tomerarnon:
Gets rid of a case of terrible type piracy we've had floating around since the beginning. 
In the process replaces all `value.()` calls with `value()`, which is probably cleaner style for our purposes.

Also supercedes #167, which brings in the fix_opt (#72) changes that are still relevant.

### DIFF
--- a/src/satisfiability/bab.jl
+++ b/src/satisfiability/bab.jl
@@ -36,16 +36,13 @@ function solve(solver::BaB, problem::Problem)
     (l_approx, l, x_l) = output_bound(solver, problem, :min)
     bound = Hyperrectangle(low = [l], high = [u])
     reach = Hyperrectangle(low = [l_approx], high = [u_approx])
-    return interpret_result(reach, bound, problem.output, x_l, x_u)
-end
 
-# This function is used by BaB and Sherlock
-function interpret_result(reach, bound, output, x_l, x_u)
-    if high(reach) < high(output) && low(reach) > low(output)
+    output = problem.output
+    if reach ⊆ output
         return ReachabilityResult(:holds, [reach])
     end
-    high(bound) > high(output)    && return CounterExampleResult(:violated, x_u)
-    low(bound)  < low(output)     && return CounterExampleResult(:violated, x_l)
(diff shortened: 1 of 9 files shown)


src/satisfiability/bab.jl
```diff
@@ -36,16 +36,13 @@ function solve(solver::BaB, problem::Problem)
     (l_approx, l, x_l) = output_bound(solver, problem, :min)
     bound = Hyperrectangle(low = [l], high = [u])
     reach = Hyperrectangle(low = [l_approx], high = [u_approx])
-    return interpret_result(reach, bound, problem.output, x_l, x_u)
-end
 
-# This function is used by BaB and Sherlock
-function interpret_result(reach, bound, output, x_l, x_u)
-    if high(reach) < high(output) && low(reach) > low(output)
+    output = problem.output
```
changliuliu, line 40: `output` is already an argument to this function.

src/satisfiability/bab.jl
```diff
@@ -99,13 +96,10 @@ end
...
-    values = Vector{Float64}(undef, 0)
-    for p in points
-        push!(values, sum(compute_output(nnet, p)))
-    end
-    value, index = ifelse(type == :min, findmin(values), findmax(values))
-    return (value, points[index])
+    points = [center(subdom), low(subdom), high(subdom)]
+    vals = [sum(compute_output(nnet, p)) for p in points]
+    i = (type == :min) ? argmin(vals) : argmax(vals)
+    return vals[i], points[i]
```
changliuliu, line 102: This is beautiful.

src/satisfiability/sherlock.jl
```diff
@@ -66,12 +73,12 @@ function local_search(problem::Problem, x::Vector{Float64}, optimizer, type::Sym
     encode_network!(model, nnet, StandardLP())
 
     gradient = get_gradient(nnet, x)
-    o = gradient * z[1]
+    o = gradient*first(z)
```
changliuliu, line 76: We need to follow SISL code standard, keep the spacing for ` * `

PR #180, opened 2020-10-06
#179 - Add new activation functions

### LINKED ISSUE #179: Add sigmoid and tanh activation functions
It would be nice to add some new types for common activation functions, such as sigmoid and tanh. See definitions [here](https://en.wikipedia.org/wiki/Activation_function).

mforets:
Closes #179.

### DIFF
--- a/src/NeuralVerification.jl
+++ b/src/NeuralVerification.jl
@@ -46,6 +46,8 @@ export
     # ReLU,
     # Max,
     # Id,
+    # Sigmoid,
+    # Tanh,
     GeneralAct,
     PiecewiseLinear,
     Problem,
--- a/src/utils/activation.jl
+++ b/src/utils/activation.jl
@@ -23,6 +23,20 @@ Identity operator
 struct Id <: ActivationFunction end
 
 """
+    Sigmoid <: ActivationFunction
+
+    (Sigmoid())(x) -> 1 ./ (1 .+ exp.(-x))
+"""
+struct Sigmoid <: ActivationFunction end
+
+"""
+    Tanh <: ActivationFunction
+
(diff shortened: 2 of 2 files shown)


tomerarnon:
Hi @mforets, I'm happy to merge this PR. It doesn't conflict with anything we have, but I wanted to let you know we have been thinking for some time about readdressing how we deal with activation functions (e.g parameterizing directly with the functions rather than types, using traits to help dispatch, etc.). At the moment, almost every algorithm we have looked at can only handle ReLUs, so haven't bothered very much in this area.

I'm curious, though, whether you needed these for anything?

mforets:
> It doesn't conflict with anything we have, but I wanted to let you know we have been thinking for some time about readdressing how we deal with activation functions

no problem. these seemed like common activation function definitions that are missing.

> I am curious though whether you needed these for anything?

yes! for sigmoid (or tanh) networks, there's [this paper](https://arxiv.org/abs/1811.01828) that proposes to compute the output of such networks using reachability analysis. we have written a proof-of-concept implementation in [NeuralNetworkAnalysis.jl](https://github.com/JuliaReach/NeuralNetworkAnalysis.jl) (see [here](https://github.com/JuliaReach/NeuralNetworkAnalysis.jl/blob/master/src/utils.jl#L75), though it's still a very WIP package).

tomerarnon:
Interesting paper! I hadn't seen it before, but the name Verisig rings a bell, so perhaps someone has already suggested it to me and I hadn't gotten around to it 🤔 

Really cool package! I poked around briefly and liked what I saw there. When I get a chance I'll also look through all of the interesting sounding examples in the docs!

PR #182, opened 2020-11-02
removing CARS workshop files

clazarus:
migrating to separate repo


tomerarnon:
What a relief! Can we add a link to the alternate repo somewhere?

clazarus:
added text and a link to the readme

PR #186, opened 2020-12-17
Implementation of NNV

castrong:
Initial implementation of NNV. It depends on [this PR in LazySets ](https://github.com/JuliaReach/LazySets.jl/pull/2132) so this shouldn't be merged until that PR is closed.

### DIFF
--- a/src/reachability/utils/reachability.jl
+++ b/src/reachability/utils/reachability.jl
@@ -22,8 +22,12 @@ end
 # these two functions can instead just be exposed without the leading "_",
 function _forward_network(solver, nnet, input)
     Z = input
-    for layer in nnet.layers
+    for (i, layer) in enumerate(nnet.layers)
+        println("forwarding layer: ", i)
         Ẑ, Z = forward_layer(solver, layer, Z)
+        if Z isa AbstractVector
+            println("   Num partitions: ", length(Z))
+        end
     end
     return Z
 end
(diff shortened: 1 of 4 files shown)


src/reachability/utils/reachability.jl
```diff
@@ -161,3 +218,36 @@ end
 
 # Define overapproximate for SymbolicIntervalMasks and SymbolicIntervalGradients
 LazySets.overapproximate(set::_SymIntOrGrad, ::Type{Hyperrectangle}) = Hyperrectangle(low=low(set), high=high(set))
+function temp_overapprox(set::Star; optimizer=GLPK.Optimizer)
```
castrong, line 221: temp_overapprox and issubset will be removed once the LazySets Star set PR finishes.

castrong:
I'm hoping for some feedback on whether to more thoroughly integrate ExactReach and NNV. I could have the same forward_act functions for both (maybe living in reachability.jl), but in this case a reader who looks at NNV.jl or ExactReach.jl wouldn't get much of a sense of what the algorithm is doing under the surface. This tradeoff might be worth it to highlight their commonalities (both just repeatedly call forward_partition to go through a ReLU layer). 

I could also potentially unify the forward_partition function to be common to both. However, we would need to decide whether the trimming used to avoid splitting on some ReLUs for NNV should be applied to ExactReach as well.

mforets:
> this PR in LazySets so this shouldn't be merged until that PR is closed.

FYI that issue is now closed and there is a [Star](https://juliareach.github.io/LazySets.jl/dev/lib/sets/Star/) type in LazySets. If you have further concerns I'd be happy to discuss here (or on the Zulip `#reachability-analysis` channel)

OTOH in https://github.com/JuliaReach/LazySets.jl/pull/2669 i'm exploring a generalization in which the predicate is a lazy intersection (say, of a polyhedron and a zonotope). That way, you can propagate such lazy intersection while storing intermediate constraints in the polyhedral part, and avoiding the (expensive) conversion of a zonotope to its hrep.

PR #188, opened 2021-04-21
These changes allow the resulting JuMP model object to encode additio…

chelseas:
…nal constraints, such as a second network, or other constraints.

### DIFF
--- a/src/optimization/utils/constraints.jl
+++ b/src/optimization/utils/constraints.jl
@@ -71,7 +71,7 @@ _ẑᵢ₊₁(m, i) = affine_map(m[:network].layers[i], m[:z][i])
 
 # Any encoding passes through here first:
 function encode_network!(model::Model, network::Network, encoding::AbstractLinearProgram)
-    get!(object_dictionary(model), :network, network)
+    object_dictionary(model)[:network] =  network # stores model[:network]
     for (i, layer) in enumerate(network.layers)
         encode_layer!(encoding, model, layer, model_params(encoding, model, i)...)
     end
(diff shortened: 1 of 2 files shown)


src/optimization/utils/constraints.jl
```diff
@@ -71,7 +71,7 @@ _ẑᵢ₊₁(m, i) = affine_map(m[:network].layers[i], m[:z][i])
 
 # Any encoding passes through here first:
 function encode_network!(model::Model, network::Network, encoding::AbstractLinearProgram)
-    get!(object_dictionary(model), :network, network)
+    object_dictionary(model)[:network] =  network # stores model[:network]
```
tomerarnon, line 74: Strictly overwriting the network here could have consequences later (if the network is needed again).

chelseas: If you are calling encode_network! and passing a network, I think it's unlikely you are wanting to reference some old stored network -- it makes sense to me that you want to reference the network you just passed. And it causes errors when you try to encode a second network, yes. 

And yes the goal is to encode two networks simultaneously. Re the second change, JuMP won't overwrite variables, it will give them distinct names, so the code that was commented "should never happen" isn't needed and throws errors when trying to encode two networks into the same jump model.

tomerarnon: I don't disagree with anything here, just wondering if this leads to problems down the line when assuming something about `model[:network]`. I'll have a look.

chelseas: Did you take a look?

PR #189, opened 2021-04-27
updated toml with version number

castrong:
Added a version number to the toml. It seems like this is needed for other packages to include it as a dependency, otherwise when trying to dev a package that depends on NeuralVerification.jl you get a message like 

ERROR: Unsatisfiable requirements detected for package NeuralVerification [146f25fa]:
 NeuralVerification [146f25fa] log:
 ├─NeuralVerification [146f25fa] has no known versions!
 └─restricted to versions * by NeuralPriorityOptimizer [7e1232c4] — no versions left
   └─NeuralPriorityOptimizer [7e1232c4] log:
     ├─possible versions are: 0.1.0 or uninstalled
     └─NeuralPriorityOptimizer [7e1232c4] is fixed to version 0.1.0

(credits to Mykel for finding this issue).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,6 @@
 name = "NeuralVerification"
 uuid = "146f25fa-00e7-11e9-3ae5-fdbac6e12fa7"
+version = "0.1.0"
 
 [deps]
 CDDLib = "3391f64e-dcde-5f30-b752-e11513730f60"


mykelk:
Yes, that's reasonable. It looks like travis is having trouble finding a compatible version of flux. Can you check into this so that the build passes?

castrong:
@mykelk the error I'm seeing in the 1.4 build may be resulting from the "complement" ambiguity that's addressed in https://github.com/sisl/NeuralVerification.jl/pull/190. Once that PR goes in, we can re-run on TRAVIS here and make sure it passes for 1.4, and 1.5.

PR #192, opened 2021-06-01
Update network.jl

mforets:
The suggested change is to facilitate reuse of some of the existing functionality but dispatch on new network implementations.

### DIFF
--- a/src/utils/network.jl
+++ b/src/utils/network.jl
@@ -1,3 +1,5 @@
+abstract type AbstractNetwork end
+
 """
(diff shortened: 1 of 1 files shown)


mforets:
Thanks. Yes, in NeuralNetworkAnalysis.jl. (For those interested, there is an upcoming JuliaCon workshop [1] where we'll introduce the package.)

[1] https://github.com/JuliaReach/JuliaCon-2021-Workshop-Its-All-Set

PR #199, opened 2021-11-12
Adding constraints so that binary variables reflect correct activation status

chelseas:
Upon inspection of optimization/utils/constraints.jl it appears that the binary variables encoding the activation status of relus in the network are not properly constrained. 
This is not a problem if you are not reasoning about the values of the binary variables, but I am using NeuralVerification.jl for an application where I am doing just this.

For example, if the lower bound of a relu is >0 the code previously just has:
```
function encode_relu(::BoundedMixedIntegerLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, l̂ᵢⱼ, ûᵢⱼ)
    if l̂ᵢⱼ >= 0.0
        @constraint(model, zᵢⱼ == ẑᵢⱼ)
...
```
And the corresponding δᵢⱼ  is free to assume either 0 or 1. 
To be most precise, δᵢⱼ should be fixed to 1 as this relu will always be active.

I have implemented these changes for the cases that I could easily understand in my own fork of this repo: 
https://github.com/chelseas/NeuralVerification.jl/tree/fix_binary

I would appreciate some assistance in making any necessary remaining changes. For example, I am using the BoundedMixedIntegerLP() type and so I have made changes focused on this and similar problem types. I'm not sure exactly how to implement this change for example, for SlackLP. 

Additionally, I'm not sure how to make these changes for the case of a layer with Identity (Id) activation. At the current moment in my own implementation, I use the init_vars() function to create the binary variables, which creates them for each layer, even if that layer were to have an identity activation. If this is going to continue to be the case, then the encode_layer for Identity activation should be changed:
```
# For an Id Layer, any encoding type defaults to this:
function encode_layer!(::AbstractLinearProgram, model::Model, layer::Layer{Id}, ẑᵢ, zᵢ, args...)
    @constraint(model, zᵢ .== ẑᵢ)
    # how do we access δᵢⱼ variable so that it can be properly constrained to 1?
    nothing
end
```
so that the corresponding binary variables is fixed to 1. However I'm not sure if it's okay to change the function signature where there is args... currently, so to explicitly require delta_ij. (Also not super familiar with the use of args... in Julia).

### DIFF
--- a/src/optimization/utils/constraints.jl
+++ b/src/optimization/utils/constraints.jl
@@ -132,7 +132,6 @@ function encode_relu(::SlackLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, sᵢⱼ)
     if δᵢⱼ
         @constraint(model, zᵢⱼ == ẑᵢⱼ + sᵢⱼ)
         @constraint(model, ẑᵢⱼ + sᵢⱼ >= 0.0)
-        # how to encode necessary constraint on δᵢⱼ here?
     else
         @constraint(model, zᵢⱼ == sᵢⱼ)
         @constraint(model, ẑᵢⱼ <= sᵢⱼ)
@@ -163,10 +162,8 @@ end
 function encode_relu(::TriangularRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, l̂ᵢⱼ, ûᵢⱼ)
     if l̂ᵢⱼ > 0.0
         @constraint(model, zᵢⱼ == ẑᵢⱼ)
-        @constraint(model, δᵢⱼ == 1)
     elseif ûᵢⱼ < 0.0
         @constraint(model, zᵢⱼ == 0.0)
-        @constraint(model, δᵢⱼ == 0)
     else
         @constraints(model, begin
                                 zᵢⱼ >= 0.0
@@ -177,18 +174,16 @@ function encode_relu(::TriangularRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, l̂ᵢⱼ
 end
 
 function encode_relu(::LinearRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ)
-    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0)) # what does this mean? how can you use ? in a constraint?
+    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0)) # in LinearRelaxedLP δᵢⱼ is a constant not a variable
 end
 
 function encode_relu(::StandardLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ)
-    if δᵢⱼ
+    if δᵢⱼ # in StandardLP δᵢⱼ is a constant, not a variable
         @constraint(model, ẑᵢⱼ >= 0.0)
         @constraint(model, zᵢⱼ == ẑᵢⱼ)
-        @constraint(model, δᵢⱼ == 1)
     else
         @constraint(model, ẑᵢⱼ <= 0.0)
         @constraint(model, zᵢⱼ == 0.0)
-        @constraint(model, δᵢⱼ == 0)
     end
 end
 
--- a/src/optimization/utils/constraints.jl
+++ b/src/optimization/utils/constraints.jl
@@ -86,6 +86,7 @@ end
 # For an Id Layer, any encoding type defaults to this:
 function encode_layer!(::AbstractLinearProgram, model::Model, layer::Layer{Id}, ẑᵢ, zᵢ, args...)
     @constraint(model, zᵢ .== ẑᵢ)
+    # how do we access δᵢⱼ variable so that it can be properly constrained to 1?
     nothing
 end
 
@@ -105,6 +106,7 @@ end
 
 
 function encode_ij(LP, model, i, j)
+    # where is this function used? Needs documentation.
     L = model[:network].layers[i]
     params = model_params(LP, model, i)
     if L.activation isa Id
@@ -124,6 +126,7 @@ function encode_relu(::SlackLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, sᵢⱼ)
     if δᵢⱼ
         @constraint(model, zᵢⱼ == ẑᵢⱼ + sᵢⱼ)
         @constraint(model, ẑᵢⱼ + sᵢⱼ >= 0.0)
+        # how to encode necessary constraint on δᵢⱼ here?
     else
         @constraint(model, zᵢⱼ == sᵢⱼ)
         @constraint(model, ẑᵢⱼ <= sᵢⱼ)
@@ -133,8 +136,10 @@ end
 function encode_relu(::BoundedMixedIntegerLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, l̂ᵢⱼ, ûᵢⱼ)
     if l̂ᵢⱼ >= 0.0
         @constraint(model, zᵢⱼ == ẑᵢⱼ)
+        @constraint(model, δᵢⱼ == 1)
     elseif ûᵢⱼ <= 0.0
         @constraint(model, zᵢⱼ == 0.0)
+        @constraint(model, δᵢⱼ == 0)
     else
         @constraints(model, begin
                                 zᵢⱼ >= 0.0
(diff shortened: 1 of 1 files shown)


src/optimization/utils/constraints.jl
```diff
@@ -164,16 +171,18 @@ function encode_relu(::TriangularRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, l̂ᵢⱼ
 end
 
 function encode_relu(::LinearRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ)
-    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0))
+    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0)) # what does this mean? how can you use ? in a constraint?
```
castrong, line 174: I agree, I am not sure what value δij will have during the creation of this constraint, or whether the @constraint macro is doing something unusual with that ternary operator.

chelseas:
I have also written this script to demonstrate the failure: 
https://gist.github.com/chelseas/a0a336f4076e1ed1e4ff87069d7b584e

src/optimization/utils/constraints.jl
tomerarnon: For the linear, triangular, and slack relaxations, `δᵢⱼ`  is not a variable of the model, but rather a concrete boolean value representing an actual (or potential) activation pattern. So in the case above, the expression could be, for example 
```julia
@constraint(model, zᵢⱼ == (true ? ẑᵢⱼ : 0.0))
# which evaluates to the following, because the expression is parsed inside-to-outside:
@constraint(model, zᵢⱼ == ẑᵢⱼ)
```
So for those two cases, there is no need (or capability) to constrain `δ`.

tomerarnon:
> For example, if the lower bound of a relu is >0 the code previously just has:
> 
> ```
> function encode_relu(::BoundedMixedIntegerLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, l̂ᵢⱼ, ûᵢⱼ)
>     if l̂ᵢⱼ >= 0.0
>         @constraint(model, zᵢⱼ == ẑᵢⱼ)
> ...
> ```
> 
> And the corresponding δᵢⱼ is free to assume either 0 or 1. To be most precise, δᵢⱼ should be fixed to 1 as this relu will always be active.

Yes, I think you are correct about this mistake!


> I would appreciate some assistance in making any necessary remaining changes. For example, I am using the BoundedMixedIntegerLP() type and so I have made changes focused on this and similar problem types. I'm not sure exactly how to implement this change for example, for SlackLP.

Actually, I think `BoundedMixedIntegerLP` is the only case where this change is necessary (since `MixedIntegerLP` forwards to the bounded version). See my reply to @castrong's comment regarding how the other encodings work for the reasoning.


>If this is going to continue to be the case, then the encode_layer for Identity activation should be changed [...] so that the corresponding binary variables is fixed to 1. However I'm not sure if it's okay to change the function signature where there is args... currently, so to explicitly require delta_ij. (Also not super familiar with the use of args... in Julia).

I think I agree here as well. And your suggested fix can be applied only to `BoundedMixedIntegerLP` as a special case rather than change this signature (unfortunately, all the different types pass through that one method, and not all of them have a `δ` variable in the model, so changing the signature could be problematic).
By the way, `args...` in a function signature in julia is like `*args` in python, i.e. a way of passing in a variable number of arguments.

src/optimization/utils/constraints.jl
chelseas: gotchya ok good to know

chelseas:
Ok good to know -- I added a separate function for ID layers for bounded mixed integer LP:

```
function encode_layer!(::BoundedMixedIntegerLP, model::Model, layer::Layer{Id}, ẑᵢ, zᵢ, δᵢⱼ, args...)
    println("Using new case!")
    @constraint(model, zᵢ .== ẑᵢ)
    @constraint(model, δᵢⱼ == 1)
    return nothing
end
```

(in the commits above)

And I think I may need to delete some constraints I added to the TriangleRelaxation types.

tomerarnon:
> And I think I may need to delete some constraints I added to the TriangleRelaxation types.

Indeed. The changes to Triangular and Standard need to be undone, and I'd prefer some of the new comments deleted as well 🙂

chelseas:
Have you checked the newest changes?
I think the comments help document what's going on. Honestly it needs more comments / documentation

src/optimization/utils/constraints.jl
```diff
@@ -103,8 +104,15 @@ function encode_layer!(SLP::SlackLP, model::Model, layer::Layer{Id}, ẑᵢ, z
     return nothing
 end
 
+# need to fix δᵢⱼ for BoundedMixedIntegerLP and possibly other types
```
tomerarnon, line 107: would rather delete this comment

src/optimization/utils/constraints.jl
```diff
@@ -103,8 +104,15 @@ function encode_layer!(SLP::SlackLP, model::Model, layer::Layer{Id}, ẑᵢ, z
...
 
+# need to fix δᵢⱼ for BoundedMixedIntegerLP and possibly other types 
+function encode_layer!(::BoundedMixedIntegerLP, model::Model, layer::Layer{Id}, ẑᵢ, zᵢ, δᵢ, args...)
+    @constraint(model, zᵢ .== ẑᵢ)
+    @constraint(model, δᵢ .== 1)
+    return nothing
+end
 
 function encode_ij(LP, model, i, j)
+    # where is this function used? Needs documentation.
```
tomerarnon, line 115: and this comment (while the statements is true that documentation is necessary, it isn't a helpful statement to find in a comment 😜)

src/optimization/utils/constraints.jl
```diff
@@ -124,6 +132,7 @@ function encode_relu(::SlackLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ, sᵢⱼ)
     if δᵢⱼ
         @constraint(model, zᵢⱼ == ẑᵢⱼ + sᵢⱼ)
         @constraint(model, ẑᵢⱼ + sᵢⱼ >= 0.0)
+        # how to encode necessary constraint on δᵢⱼ here?
```
tomerarnon, line 135: and this one

src/optimization/utils/constraints.jl
```diff
@@ -152,8 +163,10 @@ end
 function encode_relu(::TriangularRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, l̂ᵢⱼ, ûᵢⱼ)
     if l̂ᵢⱼ > 0.0
         @constraint(model, zᵢⱼ == ẑᵢⱼ)
+        @constraint(model, δᵢⱼ == 1)
```
tomerarnon, line 166: must delete (and the ==0 case below as well)

src/optimization/utils/constraints.jl
tomerarnon: This comment too.

src/optimization/utils/constraints.jl
```diff
@@ -164,16 +177,18 @@ function encode_relu(::TriangularRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, l̂ᵢⱼ
...
 function encode_relu(::LinearRelaxedLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ)
-    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0))
+    @constraint(model, zᵢⱼ == (δᵢⱼ ? ẑᵢⱼ : 0.0)) # what does this mean? how can you use ? in a constraint?
 end
 
 function encode_relu(::StandardLP, model, ẑᵢⱼ, zᵢⱼ, δᵢⱼ)
     if δᵢⱼ
         @constraint(model, ẑᵢⱼ >= 0.0)
         @constraint(model, zᵢⱼ == ẑᵢⱼ)
+        @constraint(model, δᵢⱼ == 1)
```
tomerarnon, line 187: this line (and the ==0 case) must be deleted as well.

tomerarnon (changes requested):
I believe I'm on the latest version (correct me if I'm wrong though...). I went through and noted the comments and lines that need to be removed in this review. The additions and fixes to `BoundeMixedIntegerLP` are most welcome though!

src/optimization/utils/constraints.jl
chelseas: don't see why I think it adds some much needed documentation

src/optimization/utils/constraints.jl
chelseas: Think it serves as a reminder to maintainers that they need to add documentation

src/optimization/utils/constraints.jl
chelseas: deleted this one

src/optimization/utils/constraints.jl
chelseas: ya deleted

src/optimization/utils/constraints.jl
chelseas: done

chelseas:
Can you merge, @tomerarnon ?

src/optimization/utils/constraints.jl
tomerarnon: It doesn't need to be fixed anymore, because you fixed it 😜

src/optimization/utils/constraints.jl
tomerarnon: It's not used anywhere. It seems that this was an idea for how this functionality should work that never got fleshed out all the way

tomerarnon:
Probably yes. Since CI is still down, I'm checking the branch out to test locally to make sure it all works as intended.

PR #202, opened 2022-05-23
Update util.jl

chelseas:
This fixes a bug in write_nnet  so that the input dimension is read correctly.
Due to the deprecation of the input and output bounds in the nnet header ( in general usage that I have seen)  I don't think this has caused bugs, but it is confusing because the input size is not printed correctly in the header.  I personally do look at the header for basic information about the network like input and output sizes and number of layers and I would like that to be correct.


tomerarnon:
Merged. Thanks for the fix!

PR #206, opened 2022-05-30
Added an example on how to find bounds with each solver

castrong:
I'm not sure where this should belong, but I wrote an example of how to find bounds with each solver. I'm hoping this example can help people quickly get setup with computing bounds, and would appreciate any feedback on where it or something like it can live so people can access it.


castrong:
I also put it in a github gist (as suggested by @chelseas), and we could just link that somewhere?

https://gist.github.com/castrong/c8128cd3c06554ed1999612d817efdcd

mykelk (approved):

PR #209, opened 2022-05-31
Fix/package reqs

tomerarnon:
bbcac3b767ed7fdc70b4a2e7a2970f795980c28f didn't pass the tests because something was holding the Polyhedra version back. 
From the error in CI, it seems like LazySets 1.57 uses a keyword `ztol` that was introduced in Polyhedra 0.7.0, but during the test we pull 0.6.17
Removing these two extraneous packages magically fixed it and pointed out a Flux issue. Hopefully this patches both.


castrong:
It looks like there's an error in compute_output now 

```
MethodError: no method matching zero(::Type{JLD2.ReconstructedTypes.var"##FixedPointNumbers.Normed{UInt8,8}#312"})
```
From the test in mnist_1000.jl

Perhaps JLD2's return type has changed as that's what is used to load train_x which is then reshaped and passed through the network?

tomerarnon:
Since tests pass and only documentation fails at the moment, I'll go ahead and merge.

